PR #2, opened 2015-05-18
[FIXED JENKINS-23273] /etc/sysconfig option to skip (off default) lon…

jieryn:
…g chown

### DIFF
--- a/rpm/build/SOURCES/jenkins.sysconfig.in
+++ b/rpm/build/SOURCES/jenkins.sysconfig.in
@@ -29,6 +29,16 @@ JENKINS_JAVA_CMD=""
 JENKINS_USER="@@ARTIFACTNAME@@"
 
 ## Type:        string
+## Default: "false"
+## ServiceRestart: jenkins
+#
+# Whether to skip potentially long-running chown at the
+# $JENKINS_HOME location. Do not enable this, "true", unless
+# you know what you're doing. See JENKINS-23273.
+#
+#JENKINS_INSTALL_SKIP_CHOWN="false"
+
+## Type: string
 ## Default:     "-Djava.awt.headless=true"
 ## ServiceRestart: @@ARTIFACTNAME@@
 #
--- a/rpm/build/SPECS/jenkins.spec
+++ b/rpm/build/SPECS/jenkins.spec
@@ -110,8 +110,11 @@ fi
 
 # Ensure the right ownership on files
 . /etc/sysconfig/%{name}
-chown -R ${JENKINS_USER:-%{name}} /var/log/%{name}
-chown -R ${JENKINS_USER:-%{name}} ${JENKINS_HOME:-%{workdir}}
+if test ( "x"$JENKINS_INSTALL_SKIP_CHOWN != "xtrue" ); then
+   chown -R ${JENKINS_USER:-%{name}} /var/cache/%{name}
+   chown -R ${JENKINS_USER:-%{name}} /var/log/%{name}
+   chown -R ${JENKINS_USER:-%{name}} ${JENKINS_HOME:-%{workdir}}
+fi
 
 %preun
 if [ "$1" = 0 ] ; then


daniel-beck:
Should be on by default, https://github.com/jenkinsci/packaging/blob/master/rpm/build/SOURCES/jenkins.sysconfig.in#L26 tells you to fix permissions yourself after all.

jieryn:
We already went through this in https://github.com/jenkinsci/jenkins/pull/1537 ....

daniel-beck:
> We already went through this in jenkinsci/jenkins#1537 ....

@jieryn That was about not having an option. This is about the default setting, which will affect every new installation. The expectation should be to be able to use Jenkins out of the box.

Is there a reason Jenkins `chown`s on every upgrade? I'd expect it only be necessary once after you change the user (which is when you can change this option accordingly). Which use case am I missing?

jieryn:
If you want to change the default behavior, then please open your own pull request to do so. Please stop blocking a useful change here in this PR and the several corresponding issues which already surround this PR. Please read those before holding this up. Frankly, this is getting pretty aggravating.

daniel-beck:
@jieryn I'm asking for the rationale of adding a new option for which I see no reason. There appears to be no explanation for keeping the recursive `chmod` around anywhere in this PR, PR 1537, JENKINS-12231, or JENKINS-23273 (there may well be a good reason, but so far nobody mentioned that). Therefore I still consider the current behavior to be broken, and don't think perpetuating it in an option is the correct way forward.

AFAICT, there is no recursive `chown` needed at all. As I suggested [a year ago](https://issues.jenkins-ci.org/browse/JENKINS-23273?focusedCommentId=202987&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-202987), I expect that changing this from recursive to non-recursive will be sufficient to resolve the issue of the ownership reset. The only case not covered by that would be when a user decides to reconfigure the account Jenkins runs as and forgets to `chown`, but showing a warning during installation should fix this (and I expect Jenkins to show a `BootFailure` anyway).

jieryn:
And as I suggested 6 months ago, https://github.com/jenkinsci/jenkins/pull/1537#issuecomment-70562098 , we should not muck around with default behaviors. If you want to remove it, I fully support you in some other PR. But c'mon, can we please get some improvement here? The solution in this PR is supported by the people that reported the issue, it works, and it will provide REAL relief right now. If you want to go into full on pedantic beast mode, please be my guest, but spare me having to deal with it personally.

kohsuke:
@jieryn pinged me on IRC to resolve this issue one way or the other.

My summary of the state is that there's some debate and intricacy in whether or not chmod in question should be kept, removed, changed to non-recursive, or done conditionally.

In this PR, @jieryn is suggesting an incremental improvement by providing a switch that people in the know can use. It solves his and other people's pain right away without settling the chmod debate above. (BTW I'm not sure what he thinks of this chmod, he seems to be -1 for removing chmod in [jenkins #1537](https://github.com/jenkinsci/jenkins/pull/1537) while he is now +1 in the comment above.)

@daniel-beck seems to be trying to settle the underlying debate. He is more like "hey let's agree to change chmod to non-recursive and that solves JENKINS-23273 and makes @jieryn happy." And that's a valid PoV on its own right.

Based on those fact findings, I'm going to merge this in, to provide this escape hatch immediately. [jenkins #1537](https://github.com/jenkinsci/jenkins/pull/1537) is filed half a year ago, and the frustration of @jieryn is totally understandable.

That said, on the issue of what to do with "chmod -R", I'm with @daniel-beck. I think all that's needed is to chmod directories/files managed by rpm, which are just a few. So we should make non-recursive chmod for those.

So hopefully @daniel-beck or somebody else will get to make that change and resolve JENKINS-23272 once and for all. And when that happens, `JENKINS_INSTALL_SKIP_CHOWN` switch shall be removed. Until that day, this change will provide a pain relief.

daniel-beck:
@kohsuke Is this already part of 1.617 and could be documented as such?

PR #4, opened 2015-06-15
Adding more branding variables

svanoort:
This adds parameterization for variables such as license, description, and license text, as well as project homepage, etc.

It also adds the ability to give a file path for branding info, and read branding from that file for a branding variable.


branding/common
```diff
@@ -9,3 +9,12 @@ export PORT=8080
 
 export MSI_PRODUCTCODE=415933d8-4104-47c3-aee9-66b31de07a57
 export OSX_IDPREFIX=org.jenkins-ci
+export AUTHOR="Kohsuke Kawaguchi <kk@kohsuke.org>"
+export LICENSE="MIT/X License, GPL/CDDL, ASL2"
+export HOMEPAGE=http://jenkins-ci.org/
+export CHANGELOG_PAGE=http://jenkins-ci.org/changelog
+
+export DESCRIPTION_FILE=`readlink -f description-file`
```
kohsuke: IIUC, we need something more than this is to make sure that it picks up `description-file` that's in the branding directory, not on random `pwd`

branding/README.md
```diff
@@ -14,4 +14,17 @@ Branding definition file has the following variables
 * `MSI_PRODUCTCODE`: Windows installer uses UUID to identify which MSI files are of the same lineage.
   If two MSIs have the same UUID, one will overwrite another. So if you are to produce your own MSI,
   you need to use a different UUID.
-* `OSX_IDPREFIX`: prefix of the various IDs in OS X. This follows the reverse domain name format.
\ No newline at end of file
+* `OSX_IDPREFIX`: prefix of the various IDs in OS X. This follows the reverse domain name format.
+* `AUTHOR`: author line for software
+* `LICENSE`: applicable license
+* `HOMEPAGE`: homepage for this distribution
+* `CHANGELOG_PAGE`: web page for the changelog
```
kohsuke: I'd appreciate bit more explanation of what those variables are. The idea is that somebody else should be able to write these files without hunting down where they are actually used. So the description should include the expected format and length, an example, and that sort of things.

That'll also help people decide when to reuse existing variables and when to introduce new ones.

kohsuke:
This change looks good, but waiting until after JUC to merge this.

oleg-nenashev:
IMO it also needs a hardcoded reference to Jenkins project in descriptions

PR #5, opened 2015-07-15
Avloid 'setup.mk:34: *** unterminated variable reference.  Stop.'

olivergondza:
This seems necessary for gnu make 3.81 and 4.1. Is there any public CI build I can inspect?


jtnord:
just hit this :+1:

jtnord:
> This seems necessary for gnu make 3.81 and 4.1. Is there any public CI build I can inspect?

not at this point in time.

PR #7, opened 2015-08-10
Adding noreplace to /etc/init.d/jenkins

chkpnt:
If you have adapted `/etc/init.d/jenkins` for your purposes, you don't want the file being overridden from an update.


jtnord:
:+1:  was already the case on the vanilla rpm

PR #8, opened 2015-08-10
Enable autorefresh for the new repository

chkpnt:
So the jenkins-repository gets refreshed when executing `zypper up`.


jtnord:
:+1:

PR #9, opened 2015-08-11
Fixes from testing new packaging: exports do not need quoting, and RE…

svanoort:
…ADME needs to explain passing in brand var to build package

This fixes issues discovered in testing builds for Debian: for export values, do not use quoting: this is not the shell "export" command, it is specifically for make, and handles whitespace correctly (verified). 
If you use quotes, the debian packaging will fail. 

Also, the documentation does not mention the need to supply a brand variable for initial packaging.

@reviewbybees

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,13 @@ The location of this file is set via the `WAR` variable.
 
 # Generating packages
 Run `make package` to build all the native packages.
-At minimum, you have to specify the `WAR` variable that points to the war file to be packaged.
+At minimum, you have to specify the `WAR` variable that points to the war file to be packaged and a branding file (for licensing and package descriptions). 
+You will probably need to pass in the build environment and credentials.
+
+For example:
+```shell
+make package BRAND=./branding/jenkins.mk BUILDENV=./env/test.mk CREDENTIAL=./credentials/test.mk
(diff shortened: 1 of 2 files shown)


stephenc:
:bee:

jtnord:
:bee:

svanoort:
(I should probably justify why this was merged -- it fixes something completely breaking packaging for Debian, and has been verified in testing)

oleg-nenashev:
Was there a release with the broken packaging flow? If yes, makes sense to mention it in release notes

jtnord:
@oleg-nenashev well you can't create the .deb without this change on ubuntu 15.04 without this change.
So one can only assume that @kohsuke used old versions for the last release, or that he has some weirdly different tool chain.
In either case the .deb was created so nothing worth noting.  
OT: how can we note packaging changes in the changelog now that the packaging is separate - and could be re-spun to fix packaging changes after the fact?

oleg-nenashev:
I'm OK with this clarification. 

OT: No idea. I would vote for a separate changelog for every packaging type with virtual mvn releases , but it requires additional magic in the case of the current project structure. Maybe we could just use changelog.md

PR #12, opened 2015-08-18
Clarify requirements for building on Linux

jtnord:
Specify what packages are actually needed on Linux.

@reviewbybees


tfennelly:
:bee:

oleg-nenashev:
It may become outdated easily, but at least somebody will notify us about it
:bee:

PR #13, opened 2015-08-21
Delete dead solaris build code

svanoort:
Solaris build is not supported, it's dead code (not even in the Makefile), and we should get rid of this. 

@reviewbybees


recena:
@svanoort This kind of tasks I think that it is interesting to track it in a JIRA ticket.

recena:
@svanoort Did you remove the folder?

recena:
:bee:

svanoort:
@recena Yes it does remove the directory!

PR #14, opened 2015-08-24
Fix branding script hanging if one of the branding-file.list variables is empty or unset

svanoort:
This fixes an issue with the branding script handling if the "DESCRIPTION_FILE" variable is un-set or whitespace.  Before it would hang, now it will result in an empty result.

@reviewbybees

### DIFF
--- a/bin/branding.sh
+++ b/bin/branding.sh
@@ -10,7 +10,13 @@ ARGS=()
 # Read branding files to environment variables and convert to search and replace args
 for t in $(cat $(dirname "$0")/branding-files.list);
 do
-  v="$(eval cat \$${t})"
+  # Keep going if file is not set, otherwise it'll hang on 'cat' command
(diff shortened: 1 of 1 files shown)


amuniz:
:bee: (Is the expected content of `branding-files.list` documented somewhere?)

bin/branding.sh
```diff
@@ -10,7 +10,13 @@ ARGS=()
 # Read branding files to environment variables and convert to search and replace args
 for t in $(cat $(dirname "$0")/branding-files.list);
 do
-  v="$(eval cat \$${t})"
+  # Keep going if file is not set, otherwise it'll hang on 'cat' command
+  if [ ! -z "$t" ]
```
daniel-beck: `-n`

svanoort:
Covered in review of https://github.com/jenkinsci/packaging/pull/15

PR #15, opened 2015-08-24
Fix variable substitution with @ sign in branding strings

svanoort:
This escapes the @ sign in branding variables to avoid causing incorrect results with perl substitution. 
Example of a bad string that caused issues before:   

`Kohsuke Kawaguchi <kk@kohsuke.org>`

[Builds on previous PR](https://github.com/jenkinsci/packaging/pull/14)

@reviewbybees

### DIFF
--- a/bin/branding.sh
+++ b/bin/branding.sh
@@ -17,6 +17,8 @@ do
   else
   	v="$(eval cat \$${t})"	
   fi
+  # Escape the @ signs in the variable as they will do bad things in Perl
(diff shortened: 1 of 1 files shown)


amuniz:
:bee: (sed expression tested manually :smile:)

recena:
:bee:

daniel-beck:
:bee:

svanoort:
@reviewbybees done.

svanoort:
Going ahead to merge, since this includes fix for two very irritating issues and the change is small.

PR #16, opened 2015-08-26
[JENKINS-30137] set non_interactive for ruby sftp lib.

jtnord:
The ruby sftp library will try and prompt for passwords etc which is
doomed to fail on a CI system in its default config.
Added the `non_interactive` option so that there should never be any
interaction.

@reviewbybees esp @kohsuke to make sure this won't cause a regression in his environment.


kohsuke:
:bee:

PR #18, opened 2015-09-16
Python templating replacement

svanoort:
This supplies a safer and more extensible templating mechanism than branding.sh, with tests to validate it. 

A couple useful features: 
* Compatible with branding.sh - @@VARIABLE@@ gets replaced 
* templating is safe, no worries about special characters in values breaking it (as was possible with perl/shell version)
* supports escaped strings: @@@@text@@ becomes @@text@@ (slightly clumsy, due to limits in python templating, but since @@ is unlikely to occur anyway in input... okay).  
* templating will FAIL FAST if it finds a variable that needs substitution but has no supplied value.  (Catches errors in branding definitions)
* much easier to extend/test
* very easy to drop in a different templating mechanism if desired (Jinja2, mustache, etc) - easy to test too. 

**Limitations:**
* Variable names follow python identifier restrictions.  That means case-sensitive, alphanumeric plus underscores

**New note:** I have not replaced invocations to branding.sh with branding.py yet, this is intentional, to avoid causing merge conflicts with refactoring of build scripts.

Edit: PR description has been modified significantly since initial posting to reflect latest state.  

**Below are original notions for restructuring the build (retained for comment record)**
Would it be worth pulling the content generation + templating from the actual package building?   Makes it easier to test content separate from platform-specific build bits.  Also easier to just toss the folder to platforms for final assembly. 

I was looking for an easy way to split this (DEB/SUSE/RPM are pretty much just direct folder copies), but the Mac OX X and MSI builds kind of throw a wrench in any easy approach (unless we simply split scripts into two phases).

### DIFF
--- a/bin/branding.py
+++ b/bin/branding.py
@@ -3,6 +3,7 @@
 import os
 import string
 import fnmatch
+import sys
 from optparse import OptionParser
 
 path = os.path.dirname(os.path.realpath(__file__))
@@ -16,7 +17,7 @@ with open(os.path.join(path, 'branding.list'), 'r') as branding:
 
 # Read list of environment variables from file, where each is a path to a file
 # And for each read the content of that file to the branding map
-brand-file contents from list of file name env variables and store to branding map as well
+# brand-file contents from list of file name env variables and store to branding map as well
 with open(os.path.join(path, 'branding-files.list'), 'r') as branding:
 	lines = filter(lambda x: x and not x.startswith('--'), branding.read().splitlines())
 	for line in lines:
@@ -32,10 +33,12 @@ with open(os.path.join(path, 'branding-files.list'), 'r') as branding:
 # Do search and replace using string templating for a each file
 # Throws Exceptions if I/O fails or substitution is missing vars (safety check)
 def apply_templating_to_file(path, branding_map):
-	f = open(path, "rw")
+	f = open(path, "r+")
 	f_content = string.Template(f.read()).substitute(branding_map)
+	f.seek(0)
 	f.write(f_content)
-	f.close();
+	f.truncate()
+	del f
 
 if (len(sys.argv) != 2):
 	raise Exception("Usage: branding.py [file or folder name]")
@@ -47,7 +50,8 @@ if os.path.isfile(path):
 	apply_templating_to_file(path, branding_values)
 elif os.path.isdir(path):
 	# Walks through all files in directory	
-	for root, dirnames, filename in os.walk(path):
-		apply_templating_to_file(os.path.join(dirpath, filename))
+	for root, dirnames, filenames in os.walk(path):
+		func = lambda x: apply_templating_to_file(os.path.join(root, x), branding_values)
+		map(func, filenames)
 else:
 	raise Exception("Supplied path must be file or directory")
--- a/bin/branding.py
+++ b/bin/branding.py
@@ -6,20 +6,27 @@ import fnmatch
 import sys
 from optparse import OptionParser
 
-path = os.path.dirname(os.path.realpath(__file__))
+mypath = os.path.dirname(os.path.realpath(__file__))
 branding_values = dict()
+
+def cleaned_text_lines(textcontent):
+	""" Splits file content by line boundaries, strips leading/trailing whitespace, 
+	    and removes comment and whitespace lines """
+	lines = map(lambda x: x.strip(), textcontent.splitlines())
+	lines = filter(lambda x: x and not x.startswith('#'), lines)
+	return lines
 
 # Read branding environment variable list from branding file
 # For each variable read the corresponding env variable and store to branding map
-with open(os.path.join(path, 'branding.list'), 'r') as branding:
-	lines = filter(lambda x: x and not x.startswith('--'), branding.read().splitlines())
+with open(os.path.join(mypath, 'branding.list'), 'r') as branding:
+	lines = cleaned_text_lines(branding.read())
 	branding_values = {x:os.environ.get(x) for x in lines}
 
 # Read list of environment variables from file, where each is a path to a file
 # And for each read the content of that file to the branding map
 # brand-file contents from list of file name env variables and store to branding map as well
-with open(os.path.join(path, 'branding-files.list'), 'r') as branding:
-	lines = filter(lambda x: x and not x.startswith('--'), branding.read().splitlines())
+with open(os.path.join(mypath, 'branding-files.list'), 'r') as branding:
+	lines = cleaned_text_lines(branding.read())
 	for line in lines:
 		try:
 			# Read path for this line from environment var, and open that file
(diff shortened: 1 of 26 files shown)


jtnord:
@svanoort  - I was planning on just one big "generate all the files" phase (ie do the prep work for all of the packages in one go).
Then have the individual builds do their bits on the stuff that is already transformed.

bin/branding.py
```diff
@@ -0,0 +1,57 @@
...
+# For each variable read the corresponding env variable and store to branding map
+with open(os.path.join(path, 'branding.list'), 'r') as branding:
+	lines = filter(lambda x: x and not x.startswith('--'), branding.read().splitlines())
+	branding_values = {x:os.environ.get(x) for x in lines}
+
+# Read list of environment variables from file, where each is a path to a file
+# And for each read the content of that file to the branding map
+# brand-file contents from list of file name env variables and store to branding map as well
+with open(os.path.join(path, 'branding-files.list'), 'r') as branding:
+	lines = filter(lambda x: x and not x.startswith('--'), branding.read().splitlines())
```
jtnord: # would seem more like a comment?

bin/branding.py
```diff
@@ -0,0 +1,57 @@
...
+path = os.path.dirname(os.path.realpath(__file__))
+branding_values = dict()
+
+# Read branding environment variable list from branding file
+# For each variable read the corresponding env variable and store to branding map
+with open(os.path.join(path, 'branding.list'), 'r') as branding:
+	lines = filter(lambda x: x and not x.startswith('--'), branding.read().splitlines())
+	branding_values = {x:os.environ.get(x) for x in lines}
+
+# Read list of environment variables from file, where each is a path to a file
```
jtnord: Chicken and egg problem...
You don't know where all the files are until you parse them (make's include).  All I want to know is to point at the first file that may recursively include others..

bin/branding.py
```diff
@@ -0,0 +1,57 @@
...
+		except Exception as e:
+			branding_values[line] = None
+
+# Do search and replace using string templating for a each file
+# Throws Exceptions if I/O fails or substitution is missing vars (safety check)
+def apply_templating_to_file(path, branding_map):
+	f = open(path, "r+")
+	f_content = string.Template(f.read()).substitute(branding_map)
+	f.seek(0)
+	f.write(f_content)
```
jtnord: seems a little strange.  read the file, replace the variables, then delete it?

bin/branding.py
```diff
@@ -0,0 +1,57 @@
...
+			val = f.read()	
+			branding_values[line] = val
+			f.close()
+		except Exception as e:
+			branding_values[line] = None
+
+# Do search and replace using string templating for a each file
+# Throws Exceptions if I/O fails or substitution is missing vars (safety check)
+def apply_templating_to_file(path, branding_map):
+	f = open(path, "r+")
```
jtnord: possibly better to look for files with a given extension and save them without that extension?

svanoort:
@jtnord Does that mean you're down for this templating approach so far (if so, I'll polish it up).

I +1 your notion, assuming you're not planning to lump all the files together or duplicate content.  My thought is for each build, have a platform.prep target that runs before assembly, and then we can add a 'prep-everything' make target that rolls them all up.  Being able to just prep for a single platform makes life easier for people who just want to generate a package for (for example) Ubuntu or RPM.

For deb/rpm/suse, it's basically copy/template, I think that merits being pulled out into common *and readable* code.

bin/branding.py
svanoort: I had been thinking of this too. So far it supports existing formatting, but I'll add it.

bin/branding.py
svanoort: @jtnord I don't know what you're saying here?  This is just the list that defines branding variable names to use (and the list of variables that define file content).  They don't support recursive imports or includes, and should not change often.

jtnord:
no prep per package and no make.

run one tool prepareit.py that transforms all the files for all the builds in one go.

Then you have 2 platform specific build scripts you can call (build.[sh|bat] publish.[sh|bat]) that will build and create the distribution.  
(I'm half thinking that the publish script would only create the repo bits and not push the bits, and leave that to a final script) that is run on the controlling box.
then one more cross platform script to sync all files to their final resting place on the web servers

So yes you would only be able to push the bits from one host but if you have

bin/branding.py
svanoort: In-place search and replace.  Truncate removes everything after the current point in the file.  Since we just wrote out the templated version, it's trimming off original characters that could follow that.

Might merit an explaining comment inline, I think.

bin/branding.py
svanoort: Disagree: we do all our build work in temp folders with our content.   If we keep templates and final content in the same folder, then we have to filter out the templates when rolling up packages.

bin/branding.py
jtnord: @svanoort - [https://github.com/jenkinsci/packaging/blob/master/branding/jenkins.mk] includes branding-common.  The cloudbees internal ones have even more redirection!

svanoort: @jtnord This is the variable names lists, branding.list and branding-files.list

We're talking about two different things here, this script runs after all the branding includes are evaluated, and only reads the results off the env.

svanoort:
@jtnord As promised, I have this out and ready for review.  It should now act as a drop-in replacement for branding.sh, but with several additional safety mechanisms (as called out in the description, which has been edited).  

One note: this *is* fail-fast for branding variables with no value set. I think this is desirable, but if we want to go back to replacing with empty values... it's easy to change.

jtnord:
> ONE OTHER THOUGHT Would it be worth pulling the content generation + templating from the actual package building? 
Yes - but I am working on that.

jtnord:
will this fail if I have a file to be templatized containing `@@NON_EXISTANT_VARIABLE@@`?

svanoort:
@jtnord 

> Yes - but I am working on that. 

And this continues to be a delighter for me -- this note was retained from our original discussion (I'll clarify that). 

> will this fail if I have a file to be templatized containing @@NON_EXISTANT_VARIABLE@@?

Yes, because that signifies you're trying to apply branding with no value defined... and we should fail fast in that case rather than let it pass silently until someone examines results.

jtnord:
>> | will this fail if I have a file to be templatized containing @@NON_EXISTANT_VARIABLE@@?

> Yes, because that signifies you're trying to apply branding with no value defined... and we should fail fast in that case rather than let it pass silently until someone examines results.

The unit test seems to differ - or I missed something?

```python
TEMPLATED_CONTENT = "My @@VAR@@ is @@FILECONTENT@@ and I have @@@@ESCAPES@@ and @@{NOMATCH}@@ stuff"
TEMPLATE_VARS = {'VAR': 'SPECIAL', 'FILECONTENT': 'gooooooober'}

TEMPLATE_EXPECTED = 'My SPECIAL is gooooooober and I have @@ESCAPES@@ and @@{NOMATCH}@@ stuff'
```

I would expect as `@@{NOMATCH}@@` is not defined that the script errors?

svanoort:
> I would expect as @@{NOMATCH}@@ is not defined that the script errors?

This is not super clear -- added a unit test of failed substitution and explaining for this comment.  If braces are used, this should *not* be treated as a matched variable name for substitution, for safety's sake. Only @@VARIABLE@@ is a valid substitution string.

jtnord:
:+1:

svanoort:
@reviewbybees and @olivergondza

jtnord:
:bee:

svanoort:
@reviewbybees done

PR #19, opened 2015-10-01
authenticate the CLI against the server

jtnord:
jenkins needs security and the CLI does not yet support support [ssh-agents](https://github.com/jenkinsci/jenkins/pull/920) so add a make target so that we can login and store the credentials on a CI system..

Urgh - this makes me feel very unwell.

@reviewbybees esp @kohsuke

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,9 @@ ${CLI}:
 	mv $@.tmp $@
 
 
+cliauth: ${CLI}
+	java -jar jenkins-cli.jar login --username ${JENKINS_USERNAME} --password ${JENKINS_PASSWORD}
+
 
 test.local.setup:
 	# start a test Apache server that acts as package server


jtnord:
and this doesn't work :-(

```
$ make cliauth
java -jar target/jenkins-cli.jar login --username admin22 --password bogus
"--username" is not a valid option
java -jar jenkins-cli.jar login
Saves the current credential to allow future commands to run without explicit credential information.
Makefile:82: recipe for target 'cliauth' failed
make: *** [cliauth] Error 127

jnord@cb-james /cygdrive/c/workarea/source/jenkinsci/packaging_2/packaging
```
eh?
```
$ java -jar target/jenkins-cli.jar login --username admin22 --password bogus
Bad Credentials. Search the server log for 4f024487-8d52-47a2-85ee-e3b7dc3323f4 for more details.
```

stephenc:
> "--username" is not a valid option

That error message shows up if the security realm that you are authenticating against is not an AbstractPasswordSecurityRealm...

So let's see we have [CloudbeesSecurityRealm](https://github.com/cloudbees/dac-security-plugin/blob/master/src/main/java/com/cloudbees/jenkins/plugins/dacsecurity/auth/CloudbeesSecurityRealm.java#L32) extends [OpenIdSsoSecurityRealm](https://github.com/jenkinsci/openid-plugin/blob/master/src/main/java/hudson/plugins/openid/OpenIdSsoSecurityRealm.java#L73) extends SecurityRealm... so sorry out of luck there

stephenc:
@jtnord one way would be to use the credentials binding plugin to copy the SSH private key into a file and then move it in/out of place in the make file

jtnord:
@stephenc I get that error when JENKINS_URL is set to my local jenkins instance with its internal security setup.  the same command posted at the command line works.

But thanks for the headsup about D@C looks like this is getting even more ugly as we go...  which means this is a non starter.

jtnord:
@stephenc thanks.  I forgot that there was a URL defined in the env section which was why it wasn't hitting my loclhost...

jtnord:
so non-starter so closing PR..

PR #21, opened 2015-10-02
Add the ability to specify the location of the ssh identity.

jtnord:
If "JENKINS_SSH_KEY" is set as an environment variable then use that as
the identity to pass to the Jenkins CLI.  Otherwise require the cli to
already be authenticated via jenkins-cli login as currently happens.

@reviewbybees


stephenc:
:bee:

jtnord:
@reviewbybees done

jtnord:
merging as pseudo maintainer.

PR #22, opened 2015-10-16
cp fails when copying a file in a directory contianing a space.

jtnord:
a directory may contain spaces - in which case the cp command will blow up
eg.

cp /scratch/workspace/my product/packaging/credentials/test.pkcs12 /tmp/1135/key.pkcs12
cp: target `/tmp/1135/key.pkcs12' is not a directory

@reviewbybees


svanoort:
:bee: - Apparently paths with spaces are hard to handle.

jtnord:
@reviewbybees done

PR #23, opened 2015-10-20
[FIXED JENKINS-31070] Do not dump cleartext password to the logs.

jtnord:
[JENKINS-31070](https://issues.jenkins-ci.org/browse/JENKINS-31070)

@reviewbybees

PR #24, opened 2015-10-20
[FIXED JENKINS-30280] remove dotnet installer dependency.

jtnord:
DotNet 2.0 is included in the OS or all modern systems, so don't add the
assembly - which removes the requirement for dotnet 2.0.

[JENKINS-30280](https://issues.jenkins-ci.org/browse/JENKINS-30280)

@reviewbybees


recena:
@jtnord It seems there is a conflict.

jtnord:
@recena #23 was just merged. fix is the same and a rebase force push will be coming your way.

recena:
:bee:

recena:
@jtnord If you need a tester, ping me.

jtnord:
@recena noticed an issue whilst rebasing so fix is slightly different

recena:
@jtnord Better :+1:

svanoort:
I'm not particularly familiar with MSI packaging, so let me make sure I understand correctly:

* msbuild is being run with bootstrapper.xml to generate a setup.exe for systems without dotnet 2.0 preinstalled
* This is no longer necessary, so the setup.exe does not need to be created or packaged

**Three questions:**
* Is there a reason for [build-on-jenkins.sh](https://github.com/jenkinsci/packaging/blob/57a18a5cdd24f107eed66a8c932c48944c21507d/msi/build-on-jenkins.sh) to still work with the bootstrapper.xml ? 
* Can we remove the bootstrapper.xml completely?
* Longer-term: Would it make sense to stop distributing a ZIP and instead just supply the MSI file (since all components should already be compressed)?

jtnord:
> * Is there a reason for build-on-jenkins.sh to still work with the bootstrapper.xml ?

missed it :-)

> * Can we remove the bootstrapper.xml completely?

yes

> Longer-term: Would it make sense to stop distributing a ZIP and instead just supply the MSI file (since all components should already be compressed)?

Potentially - I know some people like the zips as you see that it is obviously corrupted and/or some corporate proxies block msi/exes (most of those that do will also open the zip and inspect it so...)

I would probably just say yes here.

svanoort:
> Potentially - I know some people like the zips as you see that it is obviously corrupted and/or some corporate proxies block msi/exes (most of those that do will also open the zip and inspect it so...)

True on both points.   I have to wonder also if anyone is doing scripted installs on windows (using the setup.exe)?  Honest question, since I have no idea what is common for our windows Jenkins users.

jtnord:
> * msbuild is being run with bootstrapper.xml to generate a setup.exe for systems without dotnet 2.0 preinstalled

It creates a setup that installs .net 2.0 and then kicks the jenkins installer.

> * This is no longer necessary, so the setup.exe does not need to be created or packaged

correct

jtnord:
> *  I have to wonder also if anyone is doing scripted installs on windows

Should be using `msiexec` and the `msi` for that.

svanoort:
> Should be using msiexec and the msi for that.

True...  alright :bee: with a tiny risk of breaking people if they're doing something wacky with the setup.exe (at which point they would be be nudged towards direct consumption of the MSI anyway).

jtnord:
@reviewbybees done

svanoort:
Per discussion today, this is being held back a little bit and then rolled out to get baked in before the next LTS, for safety.

PR #27, opened 2015-10-21
[JENKINS-31074] quote arguments in osx shell

jtnord:


osx/build-on-jenkins.sh
```diff
@@ -15,10 +15,10 @@ pushd $D/src
     mv org.jenkins-ci.plis_ $OSX_IDPREFIX.plist
   popd
 popd
-$BASE/bin/branding.sh $D/src
+"$BASE/bin/branding.py" $D/src
```
jtnord: use branding.py which is much better at not needing to escape the escpaes of escapes to pass to shells at various levels.. URGGG...

svanoort:
:bee: for WIP and path handling fixex, though there are a few more places branding.sh is used (Vagrant, publish.sh, etc) that we should probably do replacement as well.

jtnord:
@reviewbybees [JENKINS-31074](https://issues.jenkins-ci.org/browse/)

oleg-nenashev:
Right link: https://issues.jenkins-ci.org/browse/JENKINS-31074

oleg-nenashev:
:bee:

svanoort:
Depends on - https://github.com/jenkinsci/packaging/pull/28 for the non-Mac builds

jtnord:
obsoletes #28

jtnord:
there are still some issues with spaces in the cwd - (especially around the use of rsync in publishing) but this is enough to unblock creating the packages.  further fixes probably require larger changes to the build which is best done as a separate change.

jtnord:
@svanoort The templated results are still as broken.  IIUC this is not a new issue - just something we found out when we switched to `branding.py` with more error handling.

jtnord:
URGH.... 
For a variable set as "export FOOBAR="  then the variable is not set to the blank value.

causes 
```
+ /home/teilo/source/jenkinsci/packaging/bin/branding.py /tmp/45192
Traceback (most recent call last):
  File "/home/teilo/source/jenkinsci/packaging/bin/branding.py", line 100, in <module>
    apply_templating_to_folder(path, branding_values)
  File "/home/teilo/source/jenkinsci/packaging/bin/branding.py", line 82, in apply_templating_to_folder
    map(func, filenames)
  File "/home/teilo/source/jenkinsci/packaging/bin/branding.py", line 81, in <lambda>
    func = lambda x: apply_templating_to_file(os.path.join(root, x), branding_map)
  File "/home/teilo/source/jenkinsci/packaging/bin/branding.py", line 72, in apply_templating_to_file
    f_content = apply_template(f.read(), branding_map)
  File "/home/teilo/source/jenkinsci/packaging/bin/branding.py", line 65, in apply_template
    return CustomTemplate(input_string).substitute(branding_map)
  File "/usr/lib/python2.7/string.py", line 176, in substitute
    return self.pattern.sub(convert, self.template)
  File "/usr/lib/python2.7/string.py", line 166, in convert
    val = mapping[named]
KeyError: 'RELEASELINE'
Makefile:67: recipe for target 'deb.publish' failed
make: *** [deb.publish] Error 1
```

jtnord:
[this line](https://github.com/jenkinsci/packaging/commit/c3708bcb7e1e8b1a11847b2174ed4d34b124a5fb#diff-a247a32c0992e3a9793a021c2c704092R79) is removing variables that are intentionally set to the empty value.

deb/publish/publish.sh
```diff
@@ -36,11 +36,11 @@ cat $D/merged/Contents | gzip -9c > $D/binary/Contents.gz
 apt-ftparchive -c $bin/release.conf release $D/binary > $D/binary/Release
 # sign the release file
 rm $D/binary/Release.gpg || true
-gpg --batch --no-use-agent --no-default-keyring --keyring $GPG_KEYRING --secret-keyring=$GPG_SECRET_KEYRING --passphrase-file $GPG_PASSPHRASE_FILE \
+gpg --batch --no-use-agent --no-default-keyring --keyring "$GPG_KEYRING" --secret-keyring="$GPG_SECRET_KEYRING" --passphrase-file "$GPG_PASSPHRASE_FILE" \
```
rsandell: Shouldn't it be `"--secret-keyring=$GPG_SECRET_KEYRING"` i.e. quotes around the whole parameter? otherwise you'd just get the quotes as part of the parameter value?

suse/publish/publish.sh
```diff
@@ -2,31 +2,31 @@
...
   # server needs 'createrepo' pacakge
-  ssh $PKGSERVER createrepo --update -o $SUSE_WEBDIR $SUSEDIR/
+  ssh $PKGSERVER createrepo --update -o "'$SUSE_WEBDIR'" "'$SUSEDIR/'"
 
   # sign the final artifact and upload the signature
-  scp $PKGSERVER:$SUSE_WEBDIR/repodata/repomd.xml repodata/
+  scp "$PKGSERVER:$SUSE_WEBDIR/repodata/repomd.xml" repodata/
 
-  gpg --batch --no-use-agent --no-default-keyring --keyring $GPG_KEYRING --secret-keyring=$GPG_SECRET_KEYRING --passphrase-file $GPG_PASSPHRASE_FILE \
+  gpg --batch --no-use-agent --no-default-keyring --keyring "$GPG_KEYRING" --secret-keyring="$GPG_SECRET_KEYRING" --passphrase-file "$GPG_PASSPHRASE_FILE" \
```
rsandell: same quote position here?

deb/publish/publish.sh
jtnord: same thing in bash... - but you logically want to quote the LHS of the equal so I find this more readable as to why this is quoted.

suse/publish/publish.sh
jtnord: same response :)

jtnord:
pull #37 (merged) fixes the issue with blank variables.

svanoort:
Have the SSH commands been tested?  I *think* they should be fine, but don't 100% trust make and its borderline insane quoting/variable rules, especially when mixing multiple quote types.   

If one's been tried, I am re-:bee: on this, I think.

jtnord:
@svanort yes this has been tested when there are no spaces in the path.  It mostly works with spaces in the paths except for the rsync but it is better than it was so IMO so good to merge quilt I will try to do a follow up to iron out the last bits.

deb/build/build.sh
```diff
@@ -13,7 +13,7 @@ mkdir -p $D
 cp -R $dir/* $D
 
 # Expand variables in the definition
-$BASE/bin/branding.sh $D/debian
+"$BASE/bin/branding.py" $D/debian
```
oleg-nenashev: This change is not a quote. Typo?

svanoort: If you mean switching from the shellscript to the python script, he's using this:  https://github.com/jenkinsci/packaging/blob/master/bin/branding.py

Something I whipped up as a safer way to apply branding (handles special characters safely, includes unit tests, and won't silently accept empty values).

abayer:
:bee:

jtnord:
@reviewbybees done

env/test.mk
```diff
@@ -9,7 +9,7 @@ export JENKINS_URL=https://cloudbees.ci.cloudbees.com/
 export PKGSERVER=${USER}@localhost
 
 # where to put binary files
-export TESTDIR=/tmp/pkg.jenkins-ci.org
+export TESTDIR==$(realpath .)/pkg.jenkins-ci.org
```
abayer: Argh, just noticed this is busted - two "="s.

svanoort: Good catch

jtnord: fixed in #41

PR #28, opened 2015-10-27
WIP for fixing issue with license branding transformation

svanoort:
Root cause of bad license file transformation is the use of fold & probably sed commands (which is legal bash, but in the make file needs to be appropriately shelled out). 

@jtnord -- any ideas?  This is the one we were talking about yesterday, I think it's stuck on a silly escape issue with the sed quoting though. (If you've got a notion, feel free to either comment or push directly to the PR branch). 

I have tested the following and it assigns correctly:  
```make
VAR:=cheese
MYVAR2:=$(shell echo $(VAR) | sed s/che/bag/g )
```

This line however raises an error with the sed:
```make
export LICENSE_TEXT_COMMENTED:=$(shell echo $(LICENSE_TEXT_COLUMN) | sed  's/^/\# /g' )
```

(I know because that's as far as it gets before errors). 

Error:
```
/bin/sh: 1: Syntax error: "(" unexpected
sed: -e expression #1, char 6: unterminated `s' command
```

My suspicion is something idiotically simple to do with special characters, but I'm puzzled because a very similar string works in a demo makefile.  Ideas?

### DIFF
--- a/setup.mk
+++ b/setup.mk
@@ -34,5 +34,5 @@ export LICENSE_TEXT_COLUMN:=$(shell fold -w 78 -s $(LICENSE_FILE))  # Format to 
 export LICENSE_TEXT_COMMENTED:=$(shell echo $(LICENSE_TEXT_COLUMN) | sed  's!^!\# !g' )
 
 # Put a dot in place of an empty line, and prepend a space
-export LICENSE_TEXT_DEB:=$(shell echo "$(LICENSE_TEXT_COLUMN)" | sed -e s!^$$!.!g -e s!^! !g )
+export LICENSE_TEXT_DEB:=$(shell echo "$(LICENSE_TEXT_COLUMN)" | sed -e 's!^$$!.!g' -e 's!^! !g' )
 
--- a/setup.mk
+++ b/setup.mk
@@ -30,9 +30,9 @@ export BASE:=$(CURDIR)
 
 # read license file and do reformatting for proper display
 export LICENSE_TEXT:=$(cat LICENSE_FILE)
-export LICENSE_TEXT_COLUMN:=$(fold -w 78 -s LICENSE_FILE)  # Format to 80 characters
-export LICENSE_TEXT_COMMENTED:=$(LICENSE_TEXT_COLUMN | sed -e 's/^/\# /g')
+export LICENSE_TEXT_COLUMN:=$(shell fold -w 78 -s $(LICENSE_FILE))  # Format to 80 characters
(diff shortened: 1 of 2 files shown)


setup.mk
```diff
@@ -30,9 +30,9 @@ export BASE:=$(CURDIR)
 
 # read license file and do reformatting for proper display
 export LICENSE_TEXT:=$(cat LICENSE_FILE)
-export LICENSE_TEXT_COLUMN:=$(fold -w 78 -s LICENSE_FILE)  # Format to 80 characters
-export LICENSE_TEXT_COMMENTED:=$(LICENSE_TEXT_COLUMN | sed -e 's/^/\# /g')
+export LICENSE_TEXT_COLUMN:=$(shell fold -w 78 -s $(LICENSE_FILE))  # Format to 80 characters
```
jtnord: on my linux box this is not actually folding the output....

setup.mk
```diff
@@ -30,9 +30,9 @@ export BASE:=$(CURDIR)
...
+export LICENSE_TEXT_COMMENTED:=$(shell echo $(LICENSE_TEXT_COLUMN) | sed  's/^/\# /g' )
```
jtnord: `/` seems to be treated as special in make?

using a different character seems to work for me `sed  's!^!\# !g'`

setup.mk
```diff
@@ -30,9 +30,9 @@ export BASE:=$(CURDIR)
...
 
 # Put a dot in place of an empty line, and prepend a space
-export LICENSE_TEXT_DEB:=$(LICENSE_TEXT_COLUMN | sed -e 's/^$/./g' -e 's/^/ /g')
+export LICENSE_TEXT_DEB:=$(shell echo "$(LICENSE_TEXT_COLUMN)" | sed -e s/^$/./g -e s/^/ /g )
```
jtnord: $ needs escaping

setup.mk
svanoort: Let me do some tinkering on this, it may be necessary to hack around this if it's misbehaving

setup.mk
svanoort: AHA!

jtnord:
I have started a POC to replace make which seems to fix the nasty not folding correctly behaviour.

svanoort:
Root cause for all this grossness turns out not to be the folds itself but
how make silently strips out newline characters all over the place :(

jtnord:
subsumed by #27

PR #29, opened 2015-10-29
[FIXED JENKINS-31302] use a more specific windows slave

jtnord:
@reviewbybees.

especially @kohsuke as he will need to reconfigure his internal cluster.

### DIFF
--- a/msi/build-on-jenkins.sh
+++ b/msi/build-on-jenkins.sh
@@ -29,10 +29,10 @@ fi
 
 case "$(uname)" in
   CYGWIN*)
-    java -jar $TARGET/jenkins-cli.jar $CLI_SSH_ARGS dist-fork -z `cygpath --dos $D/bundle.tgz` -f ${ARTIFACTNAME}.war="${WAR}" -l windows -F "${MSI}=${ARTIFACTNAME}-windows.zip" \
+    java -jar $TARGET/jenkins-cli.jar $CLI_SSH_ARGS dist-fork -z `cygpath --dos $D/bundle.tgz` -f ${ARTIFACTNAME}.war="${WAR}" -l "windows && packaging" -F "${MSI}=${ARTIFACTNAME}-windows.zip" \
           bash -ex build.sh ${ARTIFACTNAME}.war $encodedv ${ARTIFACTNAME} "${PRODUCTNAME}" ${PORT} ;;
   *)
-    java -jar $TARGET/jenkins-cli.jar $CLI_SSH_ARGS dist-fork -z $D/bundle.tgz -f ${ARTIFACTNAME}.war="${WAR}" -l windows -F "${MSI}=${ARTIFACTNAME}-windows.zip" \
+    java -jar $TARGET/jenkins-cli.jar $CLI_SSH_ARGS dist-fork -z $D/bundle.tgz -f ${ARTIFACTNAME}.war="${WAR}" -l "windows && packaging" -F "${MSI}=${ARTIFACTNAME}-windows.zip" \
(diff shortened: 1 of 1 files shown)


jtnord:
[JENKINS-31302](https://issues.jenkins-ci.org/browse/JENKINS-31302)

svanoort:
This gets a :bee: if the packaging label has been applied to the windows slaves.  One question though: is there a case where all packaging slaves are fully interchangeable?  

If not, I wonder if a better label combo would be windows && windows-packaging  (since linux, mac, and windows packaging environments are kind of special snowflakes).

jtnord:
@svanoort I don't follow.  `packaging` as a label just says this has the tools required for packaging on it - it could be a linux osx or any other slave -hence you require  "windows && packaging"
(`windows-packaging` would make `windows` redundant - yet the salve still requires windows - so you can run mainatainance tasks on all your "windows" boxes)

svanoort:
@jtnord What I mean is, windows nodes can be used as generic windows builders, as well as building windows packages.  But can they build mac/linux packages?  So while all nodes *do* packaging, and the combination of os label + packaging label will allow us to successfully ID the right slaves... the packaging label on its own isn't meaningful.  If someone created a build and just used the 'packaging' node label, what would happen?  

Or am I missing something in the way these particular slaves are set up here?

jtnord:
@svanoort 
> windows nodes can be used as generic windows builders

They should never be used as generic builders as it will screw you over without you realising it until too late as it is not a well behaved slave (java on the path - which is a big no-no in my book).

if someone just used the packaging label - they would get an arbitrary slave from the ones that had the packaging label.  (could be windows, unix, macos, solaris, a rasberry pi....) - but why would you just use the packaging label?
for instance - packaging - is just like `i386` or `x64`  - you want a 64bit windows machine you as for `x64 && windows` not `windows_x64` as then you slave labels need to be fully expanded and that is a headache to maintain.  (you would not ask for a 64bit node on its own as it does not make sense)

svanoort:
@jtnord Yeah, I'm just thinking through this and trying to think if it's the best approach for our Special Special Snowflake infrastructure here.   I mean, we can roll with it this way, to be sure. 

> They should never be used as generic builders as it will screw you over without you realising it until too late as it is not a well behaved slave (java on the path - which is a big no-no in my book).

This worries me a little.  Do we have jobs that only look for a windows slave?

> packaging - is just like i386 or x64 - you want a 64bit windows machine you as for x64 && windows not windows_x64

I agree that the windows_x64 is ridiculous - in this case we're looking at labels as orthogonal variables which each describe some different aspect of the slave, and that makes sense.  What I'm wondering is if packaging truly represents an orthogonal variable on its own, or if the real capability is what it is equipped to package for.  I mean you could have ubuntu, fedora, and say, SuSe nodes any of which can do the core linux builds (so the unique quality they share is being linux-packagers, not just packagers for their platform). 

Is this an overly pedantic point though?

tfennelly:
:bee: insofar as I get it

oleg-nenashev:
:bee:

oleg-nenashev:
@reviewbybees done

svanoort:
Dropping my comments... let's hold off merge for a little bit though, since this requires infra changes.

jtnord:
@svanoort the changes exist on cloudbees.ci and without them the build will randomly fail (if it gets a not super-soecial snowflake server).

The only infra that needs to update it @kohsuke's internal system.

jtnord:
@kohsuke reminder if you have not already you need to update your internal windows slave!

PR #30, opened 2015-10-30
Incorrect branding vars

jtnord:
@reviewbybees 

@@WAR@@ and @@HOME@@ are not branding variables - and so when using a more strict branding approach the build will fail.
As they are not branding variables they should not look and smell like branding variables.

Long term I question the sanity of doing this - as the paths being replaced can be known and created from the branding variables anyway (unless someone redefines _usr or _var macros in RPM - but that would be a WAAT...)

### DIFF
--- a/rpm/build/SOURCES/jenkins.init.in
+++ b/rpm/build/SOURCES/jenkins.init.in
@@ -31,7 +31,7 @@
 ### END INIT INFO
 
 # Check for missing binaries (stale symlinks should not happen)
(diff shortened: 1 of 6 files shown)


svanoort:
Enthusiastic :bee: 

If this builds I'd like to take it out of WIP status and merge ASAP.

jtnord:
it builds - have not actually checked the output though.

svanoort:
Merging as pseudo-maintainer, and because this fixes standing issues.

PR #34, opened 2015-12-29
Add and initially document installertest workflows

svanoort:
@abayer This is the initial hack here for OSSing the installer test flows.   I know docs are a little rough still, need to add links across the various components, but this should offer a good starting point

### DIFF
--- a/workflow/full-installertest-flow.groovy
+++ b/workflow/full-installertest-flow.groovy
@@ -1,26 +1,25 @@
 // Full installer test flow, in one file
 // You must parameterize the build with: 
 
-// @stringparameter dockerLabel (node label for docker nodes)
-// @stringparameter packagingBranch - branch to use in packaging build
+// @stringparameter dockerLabel (node label for docker nodes) - MUST be set, otherwise it'll try to run on master...
 
-// **ARTIFACTS URLS**
+// **ARTIFACTS URLS - REQUIRED**
 // Note: you can use an artifact archived in a Job build by an artifact:// URL
 //  Ex: 'artifact://full/path/to/job/buildNr#artifact.ext'
 // @stringparameter debfile URL to Debian package
 // @stringparameter rpmfile URL to CentOS/RHEL RPM package
 // @stringparameter susefile URL to SUSE RPM package
-// @stringparameter testLibraryBranch branchname to load the installertests workflow lib from
 
 // Optional build parameters
-// @stringparameter (optional) port - port number to use in testing jenkins (default 8080)
-// @stringparameter (optional) artifact (jenkins artifactname, defaults to 'jenkins')
+// @stringparameter (optional) packagingTestBranch - branch in packaging repo to use for the workflow 
+//      & the installer tests. (defaults to master)
+// @stringparameter (optional) jenkinsPort - port number to use in testing jenkins (defaults 8080)
+// @stringparameter (optional) artifactName - (jenkins artifactname, defaults to 'jenkins')
 
 // Basic parameters
-String dockerLabel = 'ubuntu-ope01'
-String packagingBranch = 'feature-docker-test-env-fixes'
-String artifactname = (artifact == null ) ? 'jenkins' : artifact
-String jenkinsPort = (port == null ) ? '8080' : "$port"
+String packagingTestBranch = (binding.hasVariable('packagingTestBranch')) ? packagingTestBranch : 'oss-dockerized-tests'
+String artifactName = (binding.hasVariable('artifactName') ? artifactName : jenkins
+String jenkinsPort = binding.hasVariable('jenkinsPort') ? jenkinsPort : '8080'
 
 // Set up
 String scriptPath = 'packaging-docker/installtests'
@@ -40,17 +39,17 @@ node(dockerLabel) {
     stage "Load Lib"
     sh 'rm -rf workflowlib'
     dir ('workflowlib') {
-        git branch: testLibraryBranch, url: 'https://github.com/jenkinsci/packaging.git'
+        git branch: packagingTestBranch, url: 'https://github.com/jenkinsci/packaging.git'
         flow = load 'workflow/installertest.groovy'
     }
     
 
     stage 'Fetch Installer'
-    flow.fetch_installers(debfile, rpmfile, susefile)
+    flow.fetchInstallers(debfile, rpmfile, susefile)
     
     sh 'rm -rf packaging-docker'
     dir('packaging-docker') {
-      git branch: packagingBranch, url: 'https://github.com/jenkinsci/packaging.git'
+      git branch: packagingTestBranch, url: 'https://github.com/jenkinsci/packaging.git'
     }
     
     // Build the sudo dockerfiles
--- a/workflow/full-installertest-flow.groovy
+++ b/workflow/full-installertest-flow.groovy
@@ -18,7 +18,7 @@
 
 // Basic parameters
 String packagingTestBranch = (binding.hasVariable('packagingTestBranch')) ? packagingTestBranch : 'oss-dockerized-tests'
-String artifactName = (binding.hasVariable('artifactName')) ? artifactName : jenkins
+String artifactName = (binding.hasVariable('artifactName')) ? artifactName : 'jenkins'
 String jenkinsPort = (binding.hasVariable('jenkinsPort')) ? jenkinsPort : '8080'
 
 // Set up
--- a/workflow/full-installertest-flow.groovy
+++ b/workflow/full-installertest-flow.groovy
@@ -23,16 +23,16 @@ String jenkinsPort = (binding.hasVariable('jenkinsPort')) ? jenkinsPort : '8080'
 
 // Set up
 String scriptPath = 'packaging-docker/installtests'
-String checkCmd = "sudo $scriptPath/service-check.sh $artifactname $jenkinsPort"
+String checkCmd = "sudo $scriptPath/service-check.sh $artifactName $jenkinsPort"
 
 // Core tests represent the basic supported linuxes, extended tests build out coverage further
 def coreTests = []
 def extendedTests = []
 coreTests[0]=["sudo-ubuntu:14.04",  ["sudo $scriptPath/debian.sh installers/deb/*.deb", checkCmd]]
-coreTests[1]=["sudo-centos:7",      ["sudo $scriptPath/centos.sh installers/rpm/*.rpm", checkCmd]]
+coreTests[1]=["sudo-centos:6",      ["sudo $scriptPath/centos.sh installers/rpm/*.rpm", checkCmd]]
 coreTests[2]=["sudo-opensuse:13.2", ["sudo $scriptPath/suse.sh installers/suse/*.rpm", checkCmd]]
 extendedTests[0]=["sudo-debian:wheezy", ["sudo $scriptPath/debian.sh installers/deb/*.deb", checkCmd]]
-extendedTests[1]=["sudo-centos:6",      ["sudo $scriptPath/centos.sh installers/rpm/*.rpm", checkCmd]]
+extendedTests[1]=["sudo-centos:7",      ["sudo $scriptPath/centos.sh installers/rpm/*.rpm", checkCmd]]
 extendedTests[2]=["sudo-ubuntu:15.10",  ["sudo $scriptPath/debian.sh installers/deb/*.deb", checkCmd]]
 
 node(dockerLabel) {
--- a/workflow/full-installertest-flow.groovy
+++ b/workflow/full-installertest-flow.groovy
@@ -1,0 +1,66 @@
+// Full installer test flow, in one file
+// You must parameterize the build with: 
+
+// @stringparameter dockerLabel (node label for docker nodes)
+// @stringparameter packagingBranch - branch to use in packaging build
+
+// **ARTIFACTS URLS**
+// Note: you can use an artifact archived in a Job build by an artifact:// URL
+//  Ex: 'artifact://full/path/to/job/buildNr#artifact.ext'
+// @stringparameter debfile URL to Debian package
+// @stringparameter rpmfile URL to CentOS/RHEL RPM package
+// @stringparameter susefile URL to SUSE RPM package
+// @stringparameter testLibraryBranch branchname to load the installertests workflow lib from
+
+// Optional build parameters
+// @stringparameter (optional) port - port number to use in testing jenkins (default 8080)
+// @stringparameter (optional) artifact (jenkins artifactname, defaults to 'jenkins')
+
+// Basic parameters
+String dockerLabel = 'ubuntu-ope01'
+String packagingBranch = 'feature-docker-test-env-fixes'
+String artifactname = (artifact == null ) ? 'jenkins' : artifact
+String jenkinsPort = (port == null ) ? '8080' : "$port"
+
+// Set up
+String scriptPath = 'packaging-docker/installtests'
+String checkCmd = "sudo $scriptPath/service-check.sh $artifactname $jenkinsPort"
+
+// Core tests represent the basic supported linuxes, extended tests build out coverage further
+def coreTests = []
+def extendedTests = []
+coreTests[0]=["sudo-ubuntu:14.04",  ["sudo $scriptPath/debian.sh installers/deb/*.deb", checkCmd]]
+coreTests[1]=["sudo-centos:7",      ["sudo $scriptPath/centos.sh installers/rpm/*.rpm", checkCmd]]
+coreTests[2]=["sudo-opensuse:13.2", ["sudo $scriptPath/suse.sh installers/suse/*.rpm", checkCmd]]
+extendedTests[0]=["sudo-debian:wheezy", ["sudo $scriptPath/debian.sh installers/deb/*.deb", checkCmd]]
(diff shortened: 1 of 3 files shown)


abayer:
Yay docs! I'll dive into this in more detail later today. Thanks!

svanoort:
@abayer Certainly!  I'll say that any shortfalls in the documentation should more than be made up by extensive comments and explanations in the flows themselves

abayer:
Definitely!

workflow/full-installertest-flow.groovy
```diff
@@ -0,0 +1,66 @@
...
+// @stringparameter rpmfile URL to CentOS/RHEL RPM package
+// @stringparameter susefile URL to SUSE RPM package
+// @stringparameter testLibraryBranch branchname to load the installertests workflow lib from
+
+// Optional build parameters
+// @stringparameter (optional) port - port number to use in testing jenkins (default 8080)
+// @stringparameter (optional) artifact (jenkins artifactname, defaults to 'jenkins')
+
+// Basic parameters
+String dockerLabel = 'ubuntu-ope01'
```
abayer: Could we change this and the `packagingBranch` to default to null, and have the flow fail out if they're still null? Makes it easier to C&P this into a Jenkinsfile, etc somewhere else...

workflow/full-installertest-flow.groovy
```diff
@@ -0,0 +1,66 @@
+// Full installer test flow, in one file
+// You must parameterize the build with: 
+
+// @stringparameter dockerLabel (node label for docker nodes)
+// @stringparameter packagingBranch - branch to use in packaging build
+
+// **ARTIFACTS URLS**
+// Note: you can use an artifact archived in a Job build by an artifact:// URL
```
abayer: I'd love it if you can optionally just specify a job/build and it figures out the artifact URLs from that, just to make life a little simpler...personal preference, but given that 99 times out of 100, `debfile`, `rpmfile` and `susefile` are all going to be under the same `artifact://full/path/to/job/buildNr` root (with an assumption that they'll all be directly under there, not in a subdirectory), auto-determination would be pleasant.

workflow/full-installertest-flow.groovy
```diff
@@ -0,0 +1,66 @@
...
+//  Ex: 'artifact://full/path/to/job/buildNr#artifact.ext'
+// @stringparameter debfile URL to Debian package
+// @stringparameter rpmfile URL to CentOS/RHEL RPM package
+// @stringparameter susefile URL to SUSE RPM package
+// @stringparameter testLibraryBranch branchname to load the installertests workflow lib from
```
abayer: Is this really needed any more? Or at least, couldn't it default to `master` (once this is merged, of course) if not specified?

workflow/full-installertest-flow.groovy
```diff
@@ -0,0 +1,66 @@
...
+String jenkinsPort = (port == null ) ? '8080' : "$port"
+
+// Set up
+String scriptPath = 'packaging-docker/installtests'
+String checkCmd = "sudo $scriptPath/service-check.sh $artifactname $jenkinsPort"
+
+// Core tests represent the basic supported linuxes, extended tests build out coverage further
+def coreTests = []
+def extendedTests = []
+coreTests[0]=["sudo-ubuntu:14.04",  ["sudo $scriptPath/debian.sh installers/deb/*.deb", checkCmd]]
```
abayer: I'm not docker-familiar enough to be sure, but should we maybe be using more distinct namespaces than just "sudo-(distro)"? Literally don't know. =)

workflow/full-installertest-flow.groovy
```diff
@@ -0,0 +1,66 @@
...
+    stage 'Build sudo dockerfiles'
+    withEnv(['HOME='+pwd()]) {
+        sh 'packaging-docker/docker/build-sudo-images.sh'
+    }
+    
+    stage 'Run Installation Tests'
+    String[] stepNames = ['install', 'servicecheck']
+    flow.execute_install_testset(coreTests, stepNames)
+    flow.execute_install_testset(extendedTests, stepNames)
+}
```
abayer: Any chance we could wrap all this in a method and return it like in `installertest.groovy`, so that, say, the `Jenkinsfile` I'm working on for `jenkins.git` can just load in the installer test without needing to copy-paste all this around and miss changes?

workflow/installertest.groovy
```diff
@@ -0,0 +1,133 @@
...
+String convertImageNameToString(String imageName, String append="") {
+    return (imageName+append).replaceAll(':','-')
+}
+
+/**
+ * Extracts the components from an {@code 'artifact://full/path/to/job/buildNr#artifact.ext'} type url.
+ * @param url the url
+ */
+@NonCPS
+def getComponentsFromArtifactUrl(String url) {
```
abayer: This is freakin' handy, and we should really be moving it into `Workflow` proper, IMO, in some way. Well, ok, actually `fetch_artifact` too. Ok, moving on, ignore me for now. =)

workflow/installertest.groovy
```diff
@@ -0,0 +1,133 @@
...
+ * 
+ * Caveats:
+ * Needs to run inside a node block, and does not stash the artifact (avoids issues with big artifacts)
+ * For smaller artifacts, you may wish to stash them
+ * @param url the url to download
+ *              {@code 'artifact://' will be downloaded via CopyArtifact build step},
+ *              Example: {@code 'artifact://full/path/to/job/buildNr#artifact.ext'}.
+ *              Anything else will be downloaded via wget
+ */
+def fetch_artifact(String url) {
```
abayer: Little nit - could we camelCase all the methods' names? Consistency and all that. =)

workflow/installertest.groovy
```diff
@@ -0,0 +1,133 @@
...
+    return (imageName+append).replaceAll(':','-')
+}
+
+/**
+ * Extracts the components from an {@code 'artifact://full/path/to/job/buildNr#artifact.ext'} type url.
+ * @param url the url
+ */
+@NonCPS
+def getComponentsFromArtifactUrl(String url) {
+    def pattern = /^artifact:\/\/([\/\w-_ ]+)\/(\d+)\/{0,1}#([\/\w\.]+)$/
```
abayer: :bug: Just noticed a problem here - the item pattern doesn't include all valid characters for a job/folder, most notably `.`.

abayer: My suggestion, from experiments - `def pattern = /^artifact:\/\/([\/\w\-\_\. ]+)\/(\d+)\/{0,1}#([\/\w\.\-\_]+)$/`. Weirdly, without escaping `-`, if it was following `\.`, it failed to match correctly, while `def pattern = /^artifact:\/\/([\/\w-_\. ]+)\/(\d+)\/{0,1}#([\/\w-_\.]+)$/` did fine. So yeah, your pick?

workflow/installertest.groovy
```diff
@@ -0,0 +1,133 @@
...
+ *              Example: {@code 'artifact://full/path/to/job/buildNr#artifact.ext'}.
+ *              Anything else will be downloaded via wget
+ */
+def fetch_artifact(String url) {
+  if (url == null) {
+    fail 'required parameter url is missing'
+  } else if (url.startsWith("artifact://")) {
+    echo "Fetching ${url} as artifact."
+    def comp = this.getComponentsFromArtifactUrl(url)
+    step([$class: 'CopyArtifact', filter: comp.artifact, projectName: comp.item, selector: [$class: 'SpecificBuildSelector', buildNumber: comp.run]])
```
abayer: this should really have `flatten:true` added, else you get the full path to the artifact.

workflow/full-installertest-flow.groovy
svanoort: I've got found a solution now for it default to master, but still allow supplying a parameter to the flow if you want to override that for custom test options (I.E. for debugging or more advanced development).   Best case solution

abayer: Sounds good.

workflow/full-installertest-flow.groovy
svanoort: Better solution:  default to master (once merged), unless you override to use another branch.  Renaming the variable anyway since the existing name is kind of unhelpful.

workflow/full-installertest-flow.groovy
svanoort: This is true: it could be a good nice-to-have in the future if others start using the flow here!

workflow/full-installertest-flow.groovy
svanoort: I did some digging and haven't really found a better way to to this (or a best practice). 
AFAICT what I'm doing here with docker containers is a new (or niche) approach without a defined best practice.

Docker registries do support namespacing, but it's generally mapped to username/org.  For example, I would be pushing personal stuff to svanoort/myimagname:tag.    One alternative is to do baseimage:sudo-version -- but I feel like that's less accurate for what we're doing here (and allows for confusion with normal base images, since the tag can be omitted).

So basically... easy to change if we find a better way :-)

svanoort:
@abayer Would you care to give this one another go?  Should be set now

abayer:
Still getting `java.net.MalformedURLException: Expected format: 'artifact://full/path/to/job/buildNr#dir/artifact.ext' but got 'artifact://2.0-workflow-poc/2.0-with-jenkinsfile/88#target/debian/jenkins_2.0_all.deb'` errors...

workflow/installertest.groovy
abayer: This still needs to be fixed. =)

workflow/installertest.groovy
abayer: Still needs to be fixed.

abayer:
With the artifact format and flattening changes added, I'm getting further than before, but still failing a ton on SuSE, a bunch of failures on Ubuntu, and one, I think, on CentOS 6.

abayer:
A thought - can we switch this to output junit-formatted test results so we can archive test results more cleanly?

workflow/full-installertest-flow.groovy
abayer: Yeah, I'd really, really like to see this.

svanoort:
@abayer The full testset is now pulled out into two library methods.   One will let you run tests with artifacts on your current node (must be in a nodeblock and have artifacts in installers/rpm/*.rpm, etc etc).   This is useful for packages build on-node.  The other one will spin up a node block and run the tests.

abayer:
Wunderbar. For the sake of argument, cc @reviewbybees but I'm already :bee:

workflow/installertest.groovy
amuniz: Perhaps @rsandell wants to include it in [pipeline-utility-steps-plugin](https://github.com/jenkinsci/pipeline-utility-steps-plugin).

svanoort: It would be a logical addition, yes... it's based on some of his goodies too :-)

workflow/installertest.groovy
```diff
@@ -0,0 +1,191 @@
...
+  // Build the sudo dockerfiles
+  stage 'Build sudo dockerfiles'
+  withEnv(['HOME='+pwd()]) {
+      sh 'packaging-docker/docker/build-sudo-images.sh'
+  }
+  
+  stage 'Run Installation Tests'
+  String[] stepNames = ['install', 'servicecheck']
+  this.executeInstallTestset(coreTests, stepNames)
+  this.executeInstallTestset(extendedTests, stepNames)
```
jglick: Seems like both these stages could just be one big `sh` script.

workflow/installertest.groovy
```diff
@@ -0,0 +1,191 @@
...
+/** Runs a series of shell commands inside a docker image
+* The output of each command set is saved to a specific file and archived
+* Errors are propagated back up the chain. 
+* @param imageName docker image name to use (must support sudo, based off the sudoable images)
+* @param shellCommands List of (string) shell commands to run within the container
+* @param stepNames List of names for each shell command step, optional
+*                    (if not supplied, then the step # will be used)
+*/
+def runShellTest(String imageName, def shellCommands, def stepNames=null) {
+  withEnv(['HOME='+pwd()]) { // Works around issues not being able to find docker install
```
jtnord: if you need to set home then this is a sign that your slave is not correctly installed.
in upstart both `HOME` and `USER` need to be defined for the slave.

workflow/installertest.groovy
```diff
@@ -0,0 +1,191 @@
...
+    def img = docker.image(imageName)
+    def fileName = convertImageNameToString(imageName,"-testOutput-")
+    img.inside() {  // Needs to be root for installation to work
```
jtnord: running as root in 2016 ho hum....

workflow/installertest.groovy
```diff
@@ -0,0 +1,191 @@
...
+          // The default posix shell does not support pipefail, so we have to invoke bash to get it
+          
+          String argument = 'bash -c \'set -o pipefail; '+cmd+" | tee \"testresults/$fileName-$name"+'.log'+'" \''
+          sh argument
+        }
+      } catch (Exception ex) {
+        archive("testresults/$fileName"+'*.log')
+        throw ex
+      }
+      archive("testresults/$fileName"+'*.log')
```
jtnord: why not just `finally` ?

workflow/installertest.groovy
```diff
@@ -0,0 +1,191 @@
...
+  def extendedTests = []
+  coreTests[0]=["sudo-ubuntu:14.04",  ["sudo $scriptPath/debian.sh installers/deb/*.deb", checkCmd]]
+  coreTests[1]=["sudo-centos:6",      ["sudo $scriptPath/centos.sh installers/rpm/*.rpm", checkCmd]]
+  coreTests[2]=["sudo-opensuse:13.2", ["sudo $scriptPath/suse.sh installers/suse/*.rpm", checkCmd]]
+  extendedTests[0]=["sudo-debian:wheezy", ["sudo $scriptPath/debian.sh installers/deb/*.deb", checkCmd]]
+  extendedTests[1]=["sudo-centos:7",      ["sudo $scriptPath/centos.sh installers/rpm/*.rpm", checkCmd]]
+  extendedTests[2]=["sudo-ubuntu:15.10",  ["sudo $scriptPath/debian.sh installers/deb/*.deb", checkCmd]]
+
+  // Run the actual work    
+  sh 'rm -rf packaging-docker'
```
jtnord: `deleteDir()` inside the `dir()`?

workflow/installertest.groovy
```diff
@@ -0,0 +1,191 @@
...
+  dir('packaging-docker') {
+    git branch: packagingTestBranch, url: 'https://github.com/jenkinsci/packaging.git'
+  }
+  
+  // Build the sudo dockerfiles
+  stage 'Build sudo dockerfiles'
+  withEnv(['HOME='+pwd()]) {
```
jtnord: see previous comment about `HOME` and `USER`

workflow/full-installertest-flow.groovy
```diff
@@ -0,0 +1,35 @@
...
+
+// Basic parameters
+String packagingTestBranch = (binding.hasVariable('packagingTestBranch')) ? packagingTestBranch : 'oss-dockerized-tests'
+String artifactName = (binding.hasVariable('artifactName')) ? artifactName : 'jenkins'
+String jenkinsPort = (binding.hasVariable('jenkinsPort')) ? jenkinsPort : '8080'
+
+
+node(dockerLabel) {
+    stage "Load Lib"
+    sh 'rm -rf workflowlib'
```
jtnord:
```
dir ('workflowlib') {
  deleteDir()
 ...
}
```

workflow/installertest.groovy
```diff
@@ -0,0 +1,191 @@
...
+      echo "Fetching ${url} as URL file."
+      sh "wget -q ${url}"
+  }
+}
+
+// Pull down the artifacts, must run in a node block
+def fetchInstallers(String debFileUrl, String rpmFileUrl, String suseFileUrl) {
+ sh 'rm -rf installers'
+ dir('installers') {
+   sh 'rm -rf deb rpm suse'
```
jtnord: seems like you just expect this to be clean so deleteDir after the dir....

workflow/installertest.groovy
rsandell: Or just fix the bugs in groovy sandbox and cps so you can use standard groovy regexp :p

workflow/installertest.groovy
svanoort: Workaround for one of the docker-workflow quirks.

workflow/installertest.groovy
svanoort: Well, *technically* running as a sudoer :-)  Perhaps I ought to make that clearer...

workflow/installertest.groovy
svanoort: I had some quirky behaviors in testing with this (nothing that I could call an out-and-out bug, but quirky).  So I left it with something I knew worked correctly.

abayer:
@jtnord So what needs to be done to get this to bee status for you? I'd really like to get it merged for https://github.com/jenkinsci/jenkins/issues/1999

workflow/installertest.groovy
jtnord: which I think was due to a badly setup salve?

svanoort: If it's more robust against slave configuration problems, why change it?

abayer:
ping @jtnord  =)

jtnord:
:bee: merge as is can be improved later.

PR #36, opened 2016-01-07
Fix artifact format, flatten copy artifacts call.

abayer:
Yes, a PR against a branch currently pending a PR. =)


svanoort:
This got lost due to a git whoopsy on my end, merging in, heh.

PR #37, opened 2016-01-19
[FIXED JENKINS-32515] allow specifying which variables may be blank.

jtnord:
It is valid for some (and only some) variables to be blank - so allow
these variables to be set to the empty string.

@reviewbybees esp @svanoort


svanoort:
:bee:  It'll do the trick for now.

jglick:
:bee:

jtnord:
@reviewbybees done
merging as pseudo code maintainer :-)

jtnord:
@recena which ID?

PR #38, opened 2016-01-21
Fix debian package not using artifact name in chowning /var/run

svanoort:
Fixes an issue with Debian packaging failing to install if:

* artifactname is *not* jenkins (example: jenkins-oc)
* /var/run/jenkins exists

Installation will fail during chown of /var/run/@@artifactname@@ because the directory does not exist.

Pending testing with installertest suite before review

### DIFF
--- a/deb/build/debian/jenkins.postinst
+++ b/deb/build/debian/jenkins.postinst
@@ -62,7 +62,7 @@ case "$1" in
 
         # older installations may use /var/run/jenkins
         # so make sure that they can delete too.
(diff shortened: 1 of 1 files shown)


svanoort:
@reviewbybees Passes basic tests

jtnord:
:bee:

jtnord:
happy as pseudo maintainer for this to be merged.

abayer:
:bee:

svanoort:
@reviewbybees done

PR #39, opened 2016-01-26
[FIXED JENKINS-31074] Quote possible space-containing paths.

abayer:
This could go further and quote all the "$D" related paths, but that's
guaranteed not to contain spaces, so I think it should be ok to skip.

cc @reviewbybees, @jtnord and @svanoort particularly.


jtnord:
are you ripping off #27 here? :-)

abayer:
oops. =)

PR #40, opened 2016-01-26
JENKINS-31074 - more attempts

abayer:
Try this one out - it seems to work for me in local tests...

cc @reviewbybees - especially @jtnord and @svanoort


jtnord:
:bee: just published some propriatary packages from a folder with spaces in their names!  Yay
:bee: :+1:

jtnord:
@reviewbybees done

PR #42, opened 2016-01-29
[JENKINS-32687] Use a configurable public key for repos.

jtnord:
This introduces a new branding variable "ORGANIZATION" that will be used
to compute the name of the key.  The key used will be taken from the
"GPG_PUBLIC_KEY" variable which needs to be configured in the credentials
make file.

@reviewbybees 

@kohsuke you will need to update your [credentials](https://github.com/jtnord/packaging/commit/d12c7e152707af955dd2ae29052010602ee31b13#diff-06e89bc08fd541c21bde330cf9ee4b1fR7) with a new variable `GPG_PUBLIC_KEY` which should point to the ascii version of your public key.

[JENKINS-32687](https://issues.jenkins-ci.org/browse/JENKINS-32687)


bin/branding.list
```diff
@@ -10,6 +10,7 @@ AUTHOR
 LICENSE
 HOMEPAGE
 CHANGELOG_PAGE
+ORGANIZATION
```
svanoort: This  should be added to the legal empty variables in python branding. 
Unless you're publishing, it's not required

jtnord: disagree - if you are publishing and its empty things get messed up!  

Lots of things are required but only for publishing (and its added to the test files that ship - so it will be set by default if you use the test branding included in this project.  (e.g RPM_URL is unused until publishing)

mslusarczyk:
as far as I understand :bee:

bin/branding.list
svanoort: While it's true that it is required for publishing... it's not actually used by @@VARIABLE@@ in the branding, it's used as a normal environment variable.  Safety checks in branding.py won't help us at all here. 

The pre-existing RPM_URL is used here in the build, for example: https://github.com/jenkinsci/packaging/blob/a610a9d26d588e7d717ff50d952d692eecfe972e/rpm/build/SOURCES/jenkins.repo

We're building out an anti-pattern here I think, long term.  If the publish fails with appropriate errors when given an invalid organization, that is probably sufficient for our needs.

test/provision/rpm.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash -ex
 /vagrant/provision/generated/common.sh
 
-wget -q -O /etc/yum.repos.d/jenkins.repo @@RPM_URL@@/@@ARTIFACTNAME@@.repo
-rpm --import @@RPM_URL@@/jenkins-ci.org.key
+wget -q -O /etc/yum.repos.d/@@ARTIFACTNAME@@.repo @@RPM_URL@@/@@ARTIFACTNAME@@.repo
```
svanoort: The actual repo file does *not* use artifact names in the repo name, in docs for installation, even for enterprise distribution. 

This needs to be in a separate PR, because we're actually changing behavior here and documentation needs to change at same time.

rpm/publish/gen.rb
```diff
@@ -21,8 +22,8 @@
 To use this repository, run the following command:
 
 <pre style="padding:0.5em; margin:1em; background:black; color:white">
-sudo wget -O /etc/yum.repos.d/jenkins.repo #{url}/#{artifactName}.repo
-sudo rpm --import #{url}/jenkins-ci.org.key
+sudo wget -O /etc/yum.repos.d/#{artifactName}.repo #{url}/#{artifactName}.repo
```
svanoort: The actual repo file does not use artifact names in the repo name, in docs for installation, even for enterprise distribution.

This needs to be in a separate PR, because we're actually changing behavior here and documentation needs to change at same time.

test/provision/rpm.sh
jtnord: well I need to change the docs for the key anyway, and it is the same documentation!

test/provision/deb.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -ex
 /vagrant/provision/generated/common.sh
 
-wget -q -O - @@DEB_URL@@/jenkins-ci.org.key | apt-key add -
+wget -q -O - @@DEB_URL@@/@@ORGANIZATION@@.key | apt-key add -
```
jtnord: Ahh this is a branding variable!!!

rpm/publish/gen.rb
jtnord: backed out change.

test/provision/rpm.sh
jtnord: backed out change

bin/branding.list
svanoort: @jtnord  Never mind: I forgot the test targets are totally isolated from the others, so we won't hit failures in branding if the variable is not set and not used.   Publish will run until hitting an error where it's used as a normal variable... rather than either failing in branding or earlier. :8ball: 

Right now in my ideal fantasyland, each of our separate build paths could run as independently as possible with minimal variables required, and fail up front if any are undefined.  But that's in the land of rainbows and unicorn farts.

svanoort:
:bee:

jtnord:
@reviewbybees done

svanoort:
I have no idea why the comment about artifact name went here, was supposed to go elsewhere *sigh.*

PR #45, opened 2016-02-18
Advance latest symlink to point to the recently published war

olivergondza:


jtnord:
:-1: breaks 3rd parties re-using these scripts

olivergondza:
@jtnord, I do not mind creating separate goal for that.

olivergondza:
I have started infra thread to fully understand what does it take to have those bits pushed and advertised properly on all around place. I am putting the on hold until then.

rtyler:
@olivergondza I believe this can, and probably should be addressed either in [sync.sh](https://github.com/jenkins-infra/jenkins-infra/blob/staging/dist/profile/files/mirrorbrain/sync.sh) or the more probable candidate: [update-latest-symlink.sh](https://github.com/jenkins-infra/jenkins-infra/blob/staging/dist/profile/files/mirrorbrain/update-latest-symlink.sh)

I wonder if this was just something that was manually done before or perhaps something I missed in migration of scripts into source control?

olivergondza:
@rtyler, This was always needed (3 LTS releases since I am involved) and there is no automation I know of to do this. I do not feel strongly where the symlink update should happen but you are right it seems to be handled by the sync tools.

Anyway, It must have run numerous times between I uploaded the war and today when I advanced the symlink manually so I think it does not work (the way I think it should). FTR, the old http://ci.jenkins-ci.org/view/Infrastructure/job/infra_mirroring was not advancing it either.

PR #47, opened 2016-02-18
Switch to sh2ju, various related tweaks.

abayer:
- change format for specifying tests so we can have distinct shell
  script files for each distro, since sh2ju uses the filenames to
  determine the test suite.
- Stop failing all over the place when tests fail - keep going so we
  can actually archive the test results, unless things blow up real good.
- Add bc and which to the sudo images to make sh2ju happy.

### DIFF
--- a/deb/build/build.sh
+++ b/deb/build/build.sh
@@ -2,6 +2,7 @@
 #
 # build a debian package from a release build
 
+hostname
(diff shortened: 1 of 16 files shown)


abayer:
cc @reviewbybees

abayer:
Replacing with a new PR.

PR #48, opened 2016-02-18
Switch to sh2ju, various related tweaks.

abayer:
- change format for specifying tests so we can have distinct shell
  script files for each distro, since sh2ju uses the filenames to
  determine the test suite.
- Stop failing all over the place when tests fail - keep going so we
  can actually archive the test results, unless things blow up real good.
- Add bc and which to the sudo images to make sh2ju happy.

cc @reviewbybees

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -11,4 +11,6 @@ RUN apt-get update \
   && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Requirement, note that v2 or later runs into issues with gem install due to Ruby 2 req
-RUN gem install net-sftp -v 1.1.1    
+RUN gem install net-sftp -v 1.1.1
+
+RUN chsh -s /bin/bash root
--- a/installtests/centos.sh
+++ b/installtests/centos.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
-set -ex  # Exit on any command failure
+set -x  # Exit on any command failure
+
+. `dirname $0`/sh2ju.sh
 
 if [ -z "$1" ]; then
   PKG_FOLDER='/tmp/packaging/target/rpm/*.rpm'
@@ -7,7 +9,11 @@ else
   PKG_FOLDER="$1"
 fi
 
-# Assume packaging is mounted to /tmp/packaging and built
-yum install -y curl
-yum install -y system-config-services java-1.7.0-openjdk  # First is b/c docker does not include service command, second is prereq
-yum -y --nogpgcheck localinstall "$PKG_FOLDER"
+dockerInstall() {
+    # Assume packaging is mounted to /tmp/packaging and built
+    yum install -y curl
+    yum install -y system-config-services java-1.7.0-openjdk  # First is b/c docker does not include service command, second is prereq
+    yum -y --nogpgcheck localinstall "$PKG_FOLDER"
+}
+
+juLog -name=centosDockerInstall dockerInstall
--- a/workflow/installertest.groovy
+++ b/workflow/installertest.groovy
@@ -205,7 +205,7 @@ def runShellTest(String imageName, def shellCommands, def stepNames=null) {
           // Second, the sh workflow step often will use the default posix shell
           // The default posix shell does not support pipefail, so we have to invoke bash to get it
           
-          String argument = cmd+" | tee \"testresults/${fileName}-${name}.log\""
+          String argument = "${cmd} | tee \"testresults/${fileName}-${name}.log\""
           sh argument
         }
       } catch (Exception ex) {
--- a/workflow/installertest.groovy
+++ b/workflow/installertest.groovy
@@ -205,7 +205,7 @@ def runShellTest(String imageName, def shellCommands, def stepNames=null) {
           // Second, the sh workflow step often will use the default posix shell
           // The default posix shell does not support pipefail, so we have to invoke bash to get it
           
-          String argument = 'bash -c \''+cmd+" | tee \"testresults/$fileName-$name"+'.log'+'" \''
+          String argument = cmd+" | tee \"testresults/${fileName}-${name}.log\""
           sh argument
         }
       } catch (Exception ex) {
--- a/workflow/installertest.groovy
+++ b/workflow/installertest.groovy
@@ -1,3 +1,116 @@
+// CONSTANTS AND CONFIG
+
+/**
+ * Gets the standard, non-platform specific test scripts and args.
+ *
+ * @param artifactName The name of the artifact we'll be testing
+ * @param jenkinsPort the port to test on.
+ *
+ * @return a list of maps of script and args to run after installation on all distros.
+ */
+private List standardScripts(String artifactName, String jenkinsPort) {
+  return [
+      [
+          "generic",
+          "service-check.sh",
+          "${artifactName} ${jenkinsPort}"
+      ]
+  ]
+}
+
+/**
+ * Gets the per-platform scripts and arguments.
+ *
+ * @return a map with platform as key and a list of script/args maps as value.
+ */
+private List platformScripts(String platform) {
+  def platScripts = [
+      [
+          "deb",
+          "debian.sh",
+          "installers/deb/*.deb"
+      ],
+      [
+          "rpm",
+          "centos.sh",
+          "installers/rpm/*.rpm"
+      ],
+      [
+          "suse",
+          "suse.sh",
+          "installers/suse/*.rpm"
+      ]
+  ]
+
+  for (int i = 0; i < platScripts.size(); i++) {
+    if (platScripts.get(i)?.get(0)?.equals(platform)) {
+      return platScripts.get(i)
+    }
+  }
+  return null
+}
+
+/**
+ * Get information on OSes to test that belong to a given category (or all)
+ *
+ * @param category Category of OSes to find - if null, returns all.
+ * @return A map with OS names as keys and information about those OSes (image name, platform and category) as values.
+ */
+private def osesToTest(String category) {
+  def osDefs = [
+      [
+          "ubuntu-14.04",
+          "sudo-ubuntu:14.04",
+          "deb",
+          "core"
+      ],
+      [
+          "ubuntu-15.04",
+          "sudo-ubuntu:15.04",
+          "deb",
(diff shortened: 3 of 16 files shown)


workflow/installertest.groovy
```diff
@@ -92,41 +205,79 @@ def runShellTest(String imageName, def shellCommands, def stepNames=null) {
           // Second, the sh workflow step often will use the default posix shell
           // The default posix shell does not support pipefail, so we have to invoke bash to get it
           
-          String argument = 'bash -c \'set -o pipefail; '+cmd+" | tee \"testresults/$fileName-$name"+'.log'+'" \''
+          String argument = 'bash -c \''+cmd+" | tee \"testresults/$fileName-$name"+'.log'+'" \''
```
jtnord: so just `cmd | tee` as you don;t need to set -o pipefail (as per comment above which needs removing!)

abayer: Ah, yeah, the `bash -c` is irrelevant.

workflow/installertest.groovy
```diff
@@ -92,41 +205,79 @@ def runShellTest(String imageName, def shellCommands, def stepNames=null) {
...
+        // scriptToRun[1] and [2] are the script name and arguments, respectively.
+        tests << shellScriptForDistro(osName, scriptPath, scriptToRun[1], scriptToRun[2])
+      }
+
+      parallelTests[osName] = {
+        try {
+          runShellTest(imgName, tests, stepNames)
+        } catch (Exception e) {
+          // Keep on trucking so we can see the full failures list
+          echo "${e}"
```
jtnord: this is dangerous.  When this gets integrated into a push installer workflow this would keep going IIUC - and the build would not stop.  You only want to run all the tests and then fail if any test fails don;t you (after grabbing the junit output)?

workflow/installertest.groovy
```diff
@@ -144,20 +295,10 @@ void runJenkinsInstallTests(String packagingTestBranch='master',
...
-  extendedTests[0]=["sudo-debian:wheezy", ["sudo $scriptPath/debian.sh installers/deb/*.deb", checkCmd]]
-  extendedTests[1]=["sudo-centos:7",      ["sudo $scriptPath/centos.sh installers/rpm/*.rpm", checkCmd]]
-  extendedTests[2]=["sudo-ubuntu:15.10",  ["sudo $scriptPath/debian.sh installers/deb/*.deb", checkCmd]]
-
-  // Run the actual work    
+  // Run the actual work
   sh 'rm -rf packaging-docker'
   dir('packaging-docker') {
-    git branch: packagingTestBranch, url: 'https://github.com/jenkinsci/packaging.git'
+    git branch: packagingTestBranch, url: 'https://github.com/abayer/packaging.git'
```
jtnord: you want to merge this with this still in?

workflow/installertest.groovy
```diff
@@ -188,4 +328,30 @@ void fetchAndRunJenkinsInstallerTest(String dockerNodeLabel, String rpmUrl, Stri
...
+ * @param distro
+ * @param scriptPath
+ * @param baseScript
+ * @param args
+ * @return the script to actually execute
+ */
+String shellScriptForDistro(String distro, String scriptPath, String baseScript, String args) {
+    String newPath = "${scriptPath}/${distro}-${baseScript}"
+    sh "cp ${scriptPath}/${baseScript} ${newPath}"
+    return "sudo ${newPath} ${args}"
```
jtnord: passwordless sudo?  is this run with docker.inside or something?

workflow/installertest.groovy
abayer: What I want is to have the build only go unstable for test failures, not go red.

workflow/installertest.groovy
abayer: ha. Fixing.

workflow/installertest.groovy
abayer: Yup.

workflow/installertest.groovy
```diff
@@ -92,41 +205,79 @@ def runShellTest(String imageName, def shellCommands, def stepNames=null) {
           // Second, the sh workflow step often will use the default posix shell
           // The default posix shell does not support pipefail, so we have to invoke bash to get it
           
-          String argument = 'bash -c \'set -o pipefail; '+cmd+" | tee \"testresults/$fileName-$name"+'.log'+'" \''
+          String argument = cmd+" | tee \"testresults/${fileName}-${name}.log\""
```
jtnord: `"${cmd} | tee 'testresults/${fileName}-${name}.log'"` is more readable no?

abayer: Yeah, probably - I started redoing the string editing without remembering the first part. Fixing!

workflow/installertest.groovy
svanoort: +1 - all that rubbish was to work around issues with passing an exit code out of a tee'd set of pipes

workflow/installertest.groovy
svanoort: With magic in the docker builds to allow it ;-)

abayer:
ping @reviewbybees, @jtnord, @svanoort - I'd like to get this in so I can re-enable installer testing in the 2.0 Jenkinsfile.

installtests/centos.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
-set -ex  # Exit on any command failure
+set -x  # Exit on any command failure
```
jtnord: -u is reccomended

installtests/centos.sh
```diff
@@ -1,5 +1,7 @@
...
+
+. `dirname $0`/sh2ju.sh
```
jtnord: $() for consitency

workflow/installertest.groovy
jtnord: Stop the build with unstable is very different to keep on trucking with an unstable result

installtests/centos.sh
abayer: `$()`? Do you mean `${}` or is `$()` something else?

installtests/centos.sh
abayer: Not being difficult, but what's `-u` do? I honestly can't remember and googling for bash command line switches is never fun. =)

workflow/installertest.groovy
abayer: I'm confused - in this scenario, at least, we don't *want* to stop the build if it's unstable, we want to keep going and make sure we run all the tests for all the platforms. Wouldn't the appropriate place for responding to the unstable and deciding to stop the build if so desired be in the push installer workflow, rather than here?

installtests/centos.sh
jtnord: fail on unset variables.

abayer: Gotcha - worthy addition. =)

installtests/centos.sh
jtnord: `$()` is the new modern backtick  - you used it earlier :)

abayer: Ah, you mean for command execution. Gotcher.

workflow/installertest.groovy
```diff
@@ -92,41 +205,79 @@ def runShellTest(String imageName, def shellCommands, def stepNames=null) {
           // Second, the sh workflow step often will use the default posix shell
           // The default posix shell does not support pipefail, so we have to invoke bash to get it
           
-          String argument = 'bash -c \'set -o pipefail; '+cmd+" | tee \"testresults/$fileName-$name"+'.log'+'" \''
+          String argument = "${cmd} | tee \"testresults/${fileName}-${name}.log\""
           sh argument
         }
       } catch (Exception ex) {
+        // And archive the test results once we're done.
+        step([$class: 'JUnitResultArchiver', keepLongStdio: true, testResults: "results/*.xml"])
```
jtnord: seems like a finally - and you catch the error above...

jtnord:
:bee: but there is a whole miss match of variables Some `$var` others `${var}`  and some commands are run with backticks - and others with `$(foo)`

workflow/installertest.groovy
abayer: I inherited this chunk of logic from what came before - I think @svanoort had hit some weird problems with `finally` vs `catch`, but I'm good either way.

abayer:
Yeah, anything I wrote myself would tend towards `${var}` and `$(foo)`, but I didn't re-do everything that came before.

docker/Dockerfile
```diff
@@ -11,4 +11,6 @@ RUN apt-get update \
   && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Requirement, note that v2 or later runs into issues with gem install due to Ruby 2 req
-RUN gem install net-sftp -v 1.1.1    
\ No newline at end of file
+RUN gem install net-sftp -v 1.1.1
+
+RUN chsh -s /bin/bash root
```
svanoort: Was this tested?   From what I can tell, pipeline does *not* respect the configured user's default shell: https://github.com/jenkinsci/durable-task-plugin/commit/3b6eb827aee03048dfe4f2cd9d2aac217f457bcf

Which calls
https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/tasks/Shell.java

I've gone through the sources, and *from what I can tell* Jenkins can set global shells but otherwise always runs /bin/sh rather than whatever your user shell is (within container or not).  

Did I miss something there?

abayer: Yeah, that probably does nothing - it was a lurking remain from experiments to deal with build failures that turned out to be due to a bad kernel version on one of ci.jenkins-ci.org's slaves. I'll remove it.

workflow/installertest.groovy
svanoort: I'm okay with it for now.  There's some room for tweaking here anyway, and I need to split off the critical from noncritical test failures.

abayer: Down the road/at some point it might be interesting to have an option to pass to fail the build for critical test failures but not for non-critical ones...

svanoort:
:bee:  I could grump about the whitespace or the juLog that could be better using test names (I think) in the repeatedly_test function, but going ahead with merge and i'll just do a follow-up PR.

PR #49, opened 2016-02-23
Catch up a few changes

abayer:
These got pushed to *my* branch, not the sh2ju branch in jenkinsci/packaging, and so didn't get merged. Oops. Here they are now in all their "glory".

cc @reviewbybees specifically @svanoort and @jtnord - sorry 'bout that!


svanoort:
:bee:  -  changes already discussed in previous PR and requested

PR #50, opened 2016-03-22
Fix sh2ju quoting

abayer:
Tests are still failing in large numbers, but that doesn't seem to be due to quotes. There's a red herring error message ("packaging-docker/installtests/sh2ju.sh: line 70: $1: unbound variable") - that really just means there are no additional arguments to the test command and we're only seeing this because we're using the "-u" flag to bash.


abayer:
cc @reviewbybees

svanoort:
:bee:  and merging.

PR #51, opened 2016-03-29
Remove AJP per [JENKINS-33776] because Jetty 9 dropped support

svanoort:
Removes AJP parameters and configuration because it has been completely dropped as of Jetty 9, and supplying AJP arguments causes winstone-Jetty to fail to start Jenkins (since the AJP connector is not supported).
 
Fixes issue [JENKINS-33776](https://issues.jenkins-ci.org/browse/JENKINS-33776).

A separate jenkins-2.0 packaging branch has been created, because pre-Jenkins 2 users may rely on AJP.


svanoort:
@reviewbybees

daniel-beck:
> A separate jenkins-2.0 packaging branch has been created, because pre-Jenkins 2 users may rely on AJP.

FYI @kohsuke

svanoort:
@daniel-beck Worth mentioning that this is a short-lived branch, only until Jenkins 2 merges to core (just so releases cut off this packaging don't lose AJP unexpectedly before that).

daniel-beck:
Right, but the RC (or maybe even \*gasp\* beta 2) needs to be from this.

jtnord:
LGTM :+1: - needs a **chuffing** big release note / warning - as many people will use this behind an Apache HTTP server - and all of a sudden their instances no longer work...
Especially as 2.0 is a drop in replacement for 1.0 :-o

andresrc:
:bee:

PR #52, opened 2016-04-08
Fix issues with 2.0 rc versions including illegal hyphens

svanoort:
This was breaking 2.0 builds, since RC flags were added without alpha/beta: https://ci.jenkins-ci.org/job/jenkins_pipeline/branch/master/7/console

RPMs are not allowed to include hyphens in versions.

### DIFF
--- a/setup.mk
+++ b/setup.mk
@@ -2,7 +2,7 @@
 export WAR?=$(error Required variable WAR must point to the jenkins.war file you are packaging)
 
 # sanitized version number
-export VERSION:=$(shell unzip -p "${WAR}" META-INF/MANIFEST.MF | grep Implementation-Version | cut -d ' ' -f2 | tr -d '\r' | sed -e "s/-SNAPSHOT//" | sed -e "s/-alpha-.*//" | sed -e "s/-beta-.*//")
+export VERSION:=$(shell unzip -p "${WAR}" META-INF/MANIFEST.MF | grep Implementation-Version | cut -d ' ' -f2 | tr -d '\r' | sed -e "s/-SNAPSHOT//" | sed -e "s/-alpha-.*//" | sed -e "s/-beta-.*//" | tr - .)
 
 # directory to place marker files for build artifacts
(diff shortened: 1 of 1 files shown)


svanoort:
@reviewbybees  and especially @abayer  and @jtnord

abayer:
Can you verify that this works with Deb packages too, and that you can upgrade from 2.0.rc.2 to 2.0?

setup.mk
```diff
@@ -2,7 +2,7 @@
 export WAR?=$(error Required variable WAR must point to the jenkins.war file you are packaging)
 
 # sanitized version number
-export VERSION:=$(shell unzip -p "${WAR}" META-INF/MANIFEST.MF | grep Implementation-Version | cut -d ' ' -f2 | tr -d '\r' | sed -e "s/-SNAPSHOT//" | sed -e "s/-alpha-.*//" | sed -e "s/-beta-.*//")
+export VERSION:=$(shell unzip -p "${WAR}" META-INF/MANIFEST.MF | grep Implementation-Version | cut -d ' ' -f2 | tr -d '\r' | sed -e "s/-SNAPSHOT//" | sed -e "s/-alpha-.*//" | sed -e "s/-beta-.*//" | tr - .)
```
jtnord: given alphas and betas are just junked - why are we keeping the rc-2 at all?

svanoort:
Dots are valid characters for Debian versioning, so no issues there (I checked).   Upgrade probably won't work, but it wouldn't have worked before (since the alpha/beta strings are stripped).

setup.mk
svanoort: I want to keep previous behavior for well defined versioning standards (alphas and betas) but this is probably a better way to do it where that's not a concern.  

Not the first time I've seen this issue hit with RPM builds.

kwhetstone:
:bee:

abayer:
:bee: I guess. =)

svanoort:
@reviewbybees done and merging

PR #54, opened 2016-05-05
Add script & helper for packaging docker image to handle running under different users

svanoort:
Attempts to create users & groups on the fly so you can use a volume mount without breaking permissions

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -12,4 +12,4 @@ RUN apt-get update \
 
 # Requirement, note that v2 or later runs into issues with gem install due to Ruby 2 req
 RUN gem install net-sftp -v 1.1.1
-
+COPY scripts/as_user.sh /bin/as_user.sh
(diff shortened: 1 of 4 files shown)


svanoort:
@reviewbybees

olivergondza:
To my understanding this require to invoke `docker run` with special arguments in order to get the account created and you have to provide ids from the outside. How can this be used from `docker.inside { ... }` step?

svanoort:
@olivergondza More or less the size of it -- in cases like this we can't really do 'inside' if we need to sudo in container.   It needs container-side configuration -- which this lands on the fly. 

If you always run under the same user you can use this as a base image and then prebake that into a custom image, but this approach lets you run under whatever user you like with sudo-ability in container.

svanoort:
Merging since this one is uncontroversial at this point (no objections, and one solid approval).

PR #55, opened 2016-06-06
Change references to pkg.jenkins-ci.org to pkg.jenkins.io.

abayer:
cc @reviewbybees, @rtyler, @kohsuke, @daniel-beck


svanoort:
Whoa keanu

recena:
LGTM 👍

env/release.mk
```diff
@@ -17,14 +17,14 @@ export RPMDIR=/srv/releases/jenkins/redhat${RELEASELINE}
 export SUSEDIR=/srv/releases/jenkins/opensuse${RELEASELINE}
 
 # where to put repository index and other web contents
-export  RPM_WEBDIR=/var/www/pkg.jenkins-ci.org.staging/redhat${RELEASELINE}
-export SUSE_WEBDIR=/var/www/pkg.jenkins-ci.org.staging/opensuse${RELEASELINE}
-export  DEB_WEBDIR=/var/www/pkg.jenkins-ci.org.staging/debian${RELEASELINE}
+export  RPM_WEBDIR=/var/www/pkg.jenkins.io.staging/redhat${RELEASELINE}
```
rtyler: This part of the changes will break things unless accompanied by a change to [this script|https://github.com/jenkins-infra/jenkins-infra/blob/staging/dist/profile/files/mirrorbrain/sync.sh#L64]

abayer: https://github.com/jenkins-infra/jenkins-infra/pull/508

svanoort:
@abayer Is this intended to only apply to pre-jenkins 2 packaging?

abayer:
Both. Lemme guess, I forgot that 2.x is on another branch. =)

svanoort:
@abayer Nah, I assumed we were just going to merge/cherry-pick stuff if both needed it, was curious though.

abayer:
Tough luck, I did a PR there too!

svanoort:
@abayer How much do we want to redirect to jenkins.io vs. jenkins-ci.org?  There are a bunch of other references:  https://github.com/jenkinsci/packaging/search?utf8=%E2%9C%93&q=jenkins-ci.org

Not sure what the impact would be of changing all them though (case-by case basis there, some might be safest to leave in place).

rtyler:
> Tough luck, I did a PR there too!

so brave

svanoort:
@rtyler That gif makes my eyes bleed! 😆

abayer:
@svanoort Piecemeal! There are probably some other places (i.e., bare references to `jenkins-ci.org`) that could/should be switched, but not everything actually has a `jenkins.io` equivalent yet. Therefore I just said "screw it, here's what's relevant for pkg, bugger everything else for now".

svanoort:
@abayer Well, I guess that makes sense -- those other usages do need to be looked at.  🐛  for not changing the .htaccess files, in the subdirectories here, for example: https://github.com/jenkinsci/packaging/tree/master/env/release

abayer:
Those refer to `mirrors.jenkins-ci.org`, not `pkg.jenkins-ci.org`. Separate change! Really, @rtyler should just go through here and change everything that's relevant. =)

svanoort:
@abayer http://mirrors.jenkins.io/ exists -- I think if we change one, we should be changing both in the same PR.  Doing it this way is a bug, because then we're tying two of our separate domains together (and make the system more brittle -- any issues in the original site will potentially hit jenkins.io). 

Whatever the opshead's version of a bug is, it's that.

rtyler:
FWIW, `pkg.jenkins-ci.org` is a CNAME of `pkg.jenkins.io` as is `mirrors.jenkins-ci.org` with regards to `mirrors.jenkins.io`

the `.io` domains became the "canonical" names when I migrated this infra to AWS a few weeks ago

svanoort:
That's somewhat better that alternatives I could see, but I still think they should be changed in step (it doesn't make sense to do otherwise).

abayer:
Happy now, @svanoort? Huh? Huh? =)

svanoort:
@abayer  I :bee: lieve I can get behind this! :+1:  (yes. :smile: )

rtyler:
So this should be merged yeah?

svanoort:
@rtyler Double checking that we don't have someone wanting to freeze packaging for a release, then yes.

rtyler:
@svanoort this code won't get execercised until @kohsuke creates a weekly release, or @olivergondza creates an RC right?

svanoort:
@rtyler Correct.  Also I just merged this into the Jenkins 2 branch as well, so we're covered there.

PR #57, opened 2016-06-10
Add a snapcraft build.

evandandrea:
Add snaps to the packaging:
- https://groups.google.com/forum/#!topic/jenkinsci-dev/3CRlaJxWSkI
- https://developer.ubuntu.com/en/desktop/

Requires Ubuntu 16.04 or later to build.


svanoort:
@evandandrea Thank you for submitting this, it is a very interesting addition. AFAIK currently our infrastructure will not support a Ubuntu 16.04 build natively, however there is an option to do this via docker containers where supported (we've got an image that can be used for that currently, but it is 14.04 based). 

CC @kohsuke and @jtnord for thoughts on adding another packaging format for Jenkins as proposed in this PR?  

My thought currently is that currently the snap format is very new and niche, so it's probably not worth the major effort to add infra support for it, test the support, and maintain it as a primary distribution in the main project.    However is someone wanted to maintain & distribute a port to snap packaging (similar to how the BSD distributions are done), that would make sense to add to the download links on Jenkins.io once it is mature.   But I'm curious what you all think...

Attaching a reference for the [snapcraft yaml format](https://developer.ubuntu.com/en/snappy/build-apps/snapcraft-syntax/).  From what I can tell this does not support many of the config options currently enabled for the other linux packages, as well as additional service integration, so it would need some work to get to the same state (and significant testing).  

**References for what a snapcraft package would need to be considered an equivalent distribution option:**
* Init scripts for Debian (showing options): https://github.com/jenkinsci/packaging/blob/master/deb/build/debian/jenkins.init
* Jenkins config args for Debian (another packaging should more or less support these options): https://github.com/jenkinsci/packaging/blob/master/deb/build/debian/jenkins.default
* Needs to  work correctly with the common plugins and options (require some significant testing)

bin/branding.list
```diff
@@ -11,6 +11,7 @@ LICENSE
 HOMEPAGE
 CHANGELOG_PAGE
 ORGANIZATION
+VERSION
```
jtnord: this appears to be very wrong...

snap/build/jenkins-config
```diff
@@ -0,0 +1,36 @@
...
+    echo "$1" >&2
+}
+
+if [ -z "$SNAP_DATA" ]; then
+    err "SNAP_DATA not set."
+    exit 1
+fi
+
+if [ ! -w "$SNAP_DATA" ]; then
+    err "Please re-run as root:"
```
jtnord: the builds will not be run as root.... :-(

snap/build/snapcraft.yaml
```diff
@@ -0,0 +1,48 @@
...
+            - usr/lib/jvm/default-java/bin
+            - usr/lib/jvm/default-java/lib
+            - usr/lib/jvm/default-java/jre/bin
+            - usr/lib/jvm/default-java/jre/lib
+            - etc/java-*
+            # Jenkins wants fonts.
+            - usr/share/fonts
+            - usr/lib/*/libfontconfig*
+            - etc/fonts/fonts.conf
+    git:
```
jtnord: why does this need git?  there is nothing in Jenkins that mandates it.

jtnord:
@svanoort I think you know my stance on the package generation (The current state needs to be sorted out before anything else is added in).

As is, this will cause breakage of the other targets (you can no longer `make package` - make is [fundamentally broken](https://bugs.launchpad.net/ubuntu/+source/make-dfsg/+bug/1531989) on later Ubuntu releases) so would need to build in parallel...

svanoort:
@evandandrea Is there a plan to address feedback here?

evandandrea:
@svanoort yes, and sorry for the extremely long delay. There have been a number of improvements to snapcraft that should make building this much easier and greatly simplify the snapcraft.yaml file. I'll get on that now.

MarkEWaite:
Based on no activity on this pull request for many years and no expressions of interest in a snap package among Jenkins users, I'm closing this pull request.  If interest arises, it can be reopened.

PR #58, opened 2016-06-21
Fix handling of zombie processes

svanoort:
Resolves the issues noted in [JENKINS-32057](https://issues.jenkins-ci.org/browse/JENKINS-32057) - root cause of that was the init script and installtests are not filtering defunct "zombie" processes.

Adds a set of tests for PID handling & correct process counts as well.

### DIFF
--- a/deb/build/debian/jenkins.init
+++ b/deb/build/debian/jenkins.init
@@ -129,7 +129,7 @@ do_start()
 # 
 get_running() 
 {
-    return `ps -U $JENKINS_USER --no-headers -f | egrep -e '(java|daemon)' | grep -c . `
+    return `ps -U $JENKINS_USER --no-headers -f | egrep -e '(java)' | grep -v defunct | grep -c . `
 }
 
 force_stop() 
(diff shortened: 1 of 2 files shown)


svanoort:
@reviewbybees

svanoort:
CC especially @abayer  and @jtnord

abayer:
:bee:

deb/build/debian/jenkins.init
```diff
@@ -129,7 +129,7 @@ do_start()
 # 
 get_running() 
 {
-    return `ps -U $JENKINS_USER --no-headers -f | egrep -e '(java|daemon)' | grep -c . `
+    return `ps -U $JENKINS_USER --no-headers -f | egrep -e '(java)' | grep -v defunct | grep -c . `
```
jtnord: why would we have defunct processes?  the launch should be detached and running as a deamon?

jtnord:
So the init script is allowing zombie processes - that seems like the underlying bug - we should not be able to create those...

deb/build/debian/jenkins.init
svanoort: It's primarily in Docker due to known issues with [Docker not being familiar with Blue Oyster Cult](https://www.youtube.com/watch?v=ClQcUyhoxTg), but can potentially happen *very briefly.*

svanoort:
@jtnord It's more that the init script doesn't disregard zombie processes when checking if a Jenkins process is still running -- normally not a major issue, just a rare minor one, but it crops up and causes a lot of noise in docker environments.

jtnord:
:bee: as I do not understand this black magic well enough to put up any counter argument

PR #59, opened 2016-06-22
[JENKINS-36143] Allow rc classifier

andresrc:
See [JENKINS-36143](https://issues.jenkins-ci.org/browse/JENKINS-36143)

@reviewbybees

### DIFF
--- a/setup.mk
+++ b/setup.mk
@@ -2,7 +2,7 @@
 export WAR?=$(error Required variable WAR must point to the jenkins.war file you are packaging)
 
 # sanitized version number
-export VERSION:=$(shell unzip -p "${WAR}" META-INF/MANIFEST.MF | grep Implementation-Version | cut -d ' ' -f2 | tr -d '\r' | sed -e "s/-SNAPSHOT//" | sed -e "s/-alpha-.*//" | sed -e "s/-beta-.*//")
+export VERSION:=$(shell unzip -p "${WAR}" META-INF/MANIFEST.MF | grep Implementation-Version | cut -d ' ' -f2 | tr -d '\r' | sed -e "s/-SNAPSHOT//" | sed -e "s/-alpha-.*//" | sed -e "s/-beta-.*//" | sed -e "s/-rc-.*//")
 
 # directory to place marker files for build artifacts
 export TARGET:=target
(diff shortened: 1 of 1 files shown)


setup.mk
```diff
@@ -35,4 +35,3 @@ export LICENSE_TEXT_COMMENTED:=$(shell echo "$(LICENSE_TEXT_COLUMN)" | sed  's!^
 
 # Put a dot in place of an empty line, and prepend a space
 export LICENSE_TEXT_DEB:=$(shell echo "$(LICENSE_TEXT_COLUMN)" | sed -e 's!^$$!.!g' -e 's!^! !g' )
-
```
jtnord: I thought keeping the trailing line is a good thing on Unix :-)

jtnord:
I'm not so sure.  What happens to the [Jenkins LTS RCs](http://mirrors.jenkins-ci.org/war-stable-rc/1.651.3/) [deb link](http://pkg.jenkins-ci.org/debian-stable-rc/)?  

I think this is OK as the Jenkins OSS seems not to be using an RC release, but a SNAPSHOT that gets published - but would like to hear some confirmation from @olivergondza before this is merged.

setup.mk
andresrc: Ups, I have to review the editor settings

andresrc:
@jtnord AFAIU that part comes from the branding. Version numbers seem to be always stripped of "decorations".

olivergondza:
We use `war.publish` goal to push the LTS RC and it would break it (as I understand the change). It would push it to non-`rc` directory.

andresrc:
The WAR directory comes from the branding and environment:
* https://github.com/jenkinsci/packaging/blob/0efc63b24b5d0807b66657c5ad61dc8d1f58af15/env/release.mk#L12
* https://github.com/jenkinsci/packaging/blob/0efc63b24b5d0807b66657c5ad61dc8d1f58af15/env/test.mk#L13

Version number decorators are stripped in the RC from the version number as well:
* http://mirrors.jenkins-ci.org/war-rc/
* http://mirrors.jenkins-ci.org/war-stable-rc/

jtnord:
@andresrc I'm pretty sure you are correct on the analysis.  the -rc lines define the location and that is specified as options to MAKE not retreived from the version.

jtnord:
:bee:

svanoort:
🐝

olivergondza:
@andresrc, @jtnord, you are right. :bee:

PR #60, opened 2016-06-27
Clean up installers during installer testing workflow

svanoort:
Does cleanup on the installer binaries when running installer testing workflow to avoid cluttering the workspace

@reviewbybees


recena:
🐝 a JIRA issue for that would be nice.

oleg-nenashev:
🐝

svanoort:
Thanks @oleg-nenashev  and @recena

PR #61, opened 2016-07-03
[JENKINS-36403] - Allow port (and potentially other ssh options) to be set/overwritten by other implementations

ctennis:


bwalding:
@reviewbybees

andresrc:
:bee:

osx/publish.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash -ex
-rsync -avz "${OSX}" "$PKGSERVER:$OSXDIR/"
+rsync -avz -e "$SSH_OPTS" "${OSX}" "$PKGSERVER:$OSXDIR/"
```
jtnord: missing a "ssh" before the SSH_OPTS?

jtnord:
:bug: has nothing to do with the referenced JIRA.

What the actual need for this is not clear wihtout a JIRA - do we really have ssh servers not running on port 22?

ctennis:
The need is purely Cloudbees, who makes use of this package by inheriting all of this stuff.  Allowing ssh access to a raw server on port 22 that is "open to the world" is against policy, so we want to dockerize the access to the package host, which necessitates running it on another port.  This patch is an effort to change as little as possible from how things work today from an outside-in perspective.

ctennis:
This should be titled JENKINS-36403, I must have gotten the wrong JIRA when making the branch.

ctennis:
Fixing bug James pointed out above, and moving to a new request with commits having the correct JENKINS JIRA.

PR #62, opened 2016-07-11
[JENKINS-36403] - Allow port (and potentially other ssh options) bee set/overwritten by other implementations

ctennis:


ctennis:
See https://github.com/jenkinsci/packaging/pull/61 for previous attempt at a pull request.

@reviewbybees

andresrc:
:bee:

jtnord:
:bee: - but this has the possibility to break other users of these scripts as the script will now fail if they use a custom environment file.

ctennis:
I think it will be okay right?  If they have a custom file where SSH_OPTS doesn't evaluate to anything, everything will continue working as before.

jtnord:
> If they have a custom file where SSH_OPTS doesn't evaluate to anything, everything will continue working as before.

Yes - my point is the build would break until they added `export SSH_OPTS=""`  (which is a really simple fix, but that you probably want to add this to any internal evenvironments you may have before this is merged :-) so you do not break your own internal builds :-P )

andresrc:
@reviewbybees done

@olivergondza looks ok for you?

olivergondza:
:+1:

PR #64, opened 2016-08-23
[JENKINS-36403] Fixes following newly added SSH_OPTS env variable

armfergom:
There were changes made in #62 to allow custom ssh ports for the packaging scripts. 

The bit included in this PR was missed and causes errors when using a port different from 22.

- [x] Fix in ruby script to use correct port
- [x] Use scp ops since they are not equivalent to ssh ops

@reviewbybees esp. @jtnord

### DIFF
--- a/bin/list-packages.rb
+++ b/bin/list-packages.rb
@@ -8,8 +8,12 @@ require 'net/sftp' # gem install net-sftp
 def list_packages(path,glob,rel)
 	# user name and server name
 	u,s=ENV['PKGSERVER'].split("@")
+        # Look for port option
+        portOption = ENV['SSH_OPTS'].scan(/-p [0-9]+/)
+        # Get port if set
+        p = portOption.empty? ? 22 : portOption.first.split(" ").last
 
-    Net::SFTP.start(s,u) do |sftp|
+    Net::SFTP.start(s,u,:port=>p) do |sftp|
         debs = [];
         sftp.dir.glob(path,glob) { |f| debs << f }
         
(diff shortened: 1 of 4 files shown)


alvarolobato:
:bee:

jtnord:
:bug:  SSH_OPTS can be used for any arbitrary ssh parameters.  There is nothing to say that it only contains a port, or that the port must be the first option.

bin/list-packages.rb
```diff
@@ -8,8 +8,10 @@
 def list_packages(path,glob,rel)
 	# user name and server name
 	u,s=ENV['PKGSERVER'].split("@")
+        # port
+        p=ENV['SSH_OPTS'].empty? ? 22 : ENV['SSH_OPTS'].split(" ").last
```
jtnord: `SSH_OPTS= -o StrictHostKeyChecking=no`?

armfergom:
@jtnord fair enough. Better now?

jtnord:
@armfergom  :bee:

armfergom:
Brilliant. Would it be possible to get it merged?

jtnord:
@reviewbybees done

jtnord:
@olivergondza any comments?

olivergondza:
Looks good to me as I understand it.

armfergom:
There is at least one more error in the scripts. Specifically, `SSH_OPTS `is used in `scp` command but the options available are not equivalent for both commands. A new variable,`SCP_OPTS` has been added.

jtnord:
:bee: seems reasonable even if this is now stating to set off my spider senses....

alvarolobato:
:bee:

recampbell:
@jtnord can you merge?

PR #65, opened 2016-08-26
[JENKINS-37721] the service in the jenkins.xml must match what wix installed

jtnord:
Make sure that the service name used by `jenkins.exe` is the same as the
service name registered by the msi.

Not an issue for vanilla Jenkins as the registry is case insensitive and
the fact that jenkins != Jenkins is ok, however when the artifact name
contains a dash such as jenkins-oc and the camel artifact name is
forbidden to cotain dashes such that they artifact name is not jusy a
lower case equivallent of camelartifactname then things do not work.

Whilst I was there I also corrected the description of the service in the
XML - ot that it is used , but more because it looks better (and someone
could use this to install the service if they did not want to use the MSI
by copying the files and running `jenkins.exe install`

@reviewbybees - this internally maps to CJP-4641

### DIFF
--- a/msi/build.sh
+++ b/msi/build.sh
@@ -6,8 +6,9 @@ ENCODEDVERSION="$2"
 ARTIFACTNAME="$3"
 PRODUCTNAME="$4"
 PORT="$5"
-if [ "" == "$PORT" ]; then
-  echo "build.sh path/to/jenkins.war version artifactName port"
+SERVICENAME="$6"
+if [ "" == "$SERVICENAME" ]; then
+  echo "build.sh path/to/jenkins.war version artifactName port servicename"
   exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


jtnord:
@andresrc @armfergom merged to ssh_opts_plus_jenkins-37721  for testing.

svanoort:
Yeah okay that looks reasonable to me 🐝  and 👍

jtnord:
@reviewbybees done.

jtnord:
@andresrc ping me when the testing has verified this and I will merge.

andresrc:
@jtnord tested. This and https://github.com/jenkinsci/packaging/pull/64 are ready to merge.

msi/build.sh
```diff
@@ -6,18 +6,35 @@ ENCODEDVERSION="$2"
 ARTIFACTNAME="$3"
 PRODUCTNAME="$4"
 PORT="$5"
-if [ "" == "$PORT" ]; then
-  echo "build.sh path/to/jenkins.war version artifactName port"
+SERVICENAME="$6"
+if [ "" == "$SERVICENAME" ]; then
```
oleg-nenashev: Generally this change was not an equivalent if one of parameters is specified. Likely NIT

PR #66, opened 2016-08-27
Fix check_tcp_port in debian init script

br0ch0n:
I don't know if this function still needs to exist at all, but:

In its current form, Jenkins will fail to start if you have
multiple interfaces on your box (or even just public + localhost)
listening on the same port, e.g. 8080
Per the Jenkins docs, we should use the var $HTTP_HOST along
with --httpListenAddress for binding to explicit interfaces.
The init script now takes this into account when checking port
availability.

### DIFF
--- a/deb/build/debian/jenkins.init
+++ b/deb/build/debian/jenkins.init
@@ -71,18 +71,26 @@ check_tcp_port() {
     local service=$1
     local assigned=$2
     local default=$3
+    local assigned_address=$4
+    local default_address=$5
 
-    if [ -n "$assigned" ]; then 
+    if [ -n "$assigned" ]; then
         port=$assigned
     else
         port=$default
     fi
-    
-    count=`netstat --listen --numeric-ports | grep \:$port[[:space:]] | grep -c . `
+
+    if [ -n "$assigned_address" ]; then
+        address=$assigned_address
+    else
+        address=$default_address
+    fi
+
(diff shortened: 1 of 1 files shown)


deb/build/debian/jenkins.init
```diff
@@ -103,7 +111,7 @@ do_start()
 
     # Verify that the jenkins port is not already in use, winstone does not exit
     # even for BindException
-    check_tcp_port "http" "$HTTP_PORT" "@@PORT@@" || return 2
+    check_tcp_port "http" "$HTTP_PORT" "@@PORT@@" "$HTTP_HOST" "0.0.0.0" || return 2
```
br0ch0n: > It doesn't make sense to force the check on all interfaces here.

That was the whole point of this PR. The previous behavior of init was checking all interfaces with no option to change it (again I don't know if this check is really needed at all). Now it only checks 0.0.0.0 if you haven't overridden HTTP_HOST in /etc/default/jenkins.  I added 0.0.0.0 as the default fallback here since that's what jenkins is going to be listening on anyway if you aren't using HTTP_HOST.  I'll submit a comment change in jenkins.default to make it clearer that one should do:
```
JENKINS_ARGS="--webroot=/var/cache/$NAME/war --httpPort=$HTTP_PORT --httpListenA
ddress=$HTTP_HOST"
```
if they want to bind to a certain interface and have things work in concert with the init script.  HTTP_HOST is the var we should be using per https://wiki.jenkins-ci.org/display/JENKINS/Starting+and+Accessing+Jenkins

PR #67, opened 2016-09-05
Create dedicated build user for packaging container

olivergondza:
Add dedicated build users with configurable gid and uid. Otherwise, when used from `docker.build( ... ).inside { ... }`, the build is running with artificial uid/gid combination without proper environment configured causing tools to fail.

https://groups.google.com/forum/#!topic/jenkinsci-dev/BHGAEcQK7UE

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,15 @@
 FROM ubuntu:14.04
 MAINTAINER svanoort <samvanoort@gmail.com>
 
+# User id of the build user
+ARG uid=4242
+ENV uid $uid
+
+# Group id of the build user
+ARG gid=4242
+ENV gid $gid
+
+RUN groupadd builder -g $gid && useradd builder -u $uid -g $gid -m -d /home/builder
 
 # Dependencies plus vim for convenience, wget if you're trying to get fancy
 RUN apt-get update \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,20 +2,20 @@ FROM ubuntu:14.04
 MAINTAINER svanoort <samvanoort@gmail.com>
 
 # User id of the build user
-ARG uid=4242
+ARG uid=1000
 ENV uid $uid
 
 # Group id of the build user
-ARG gid=4242
+ARG gid=1000
 ENV gid $gid
 
-RUN groupadd builder -g $gid && useradd builder -u $uid -g $gid -m -d /home/builder
+RUN groupadd jenkins -g $gid && useradd jenkins -u $uid -g $gid -m -d /home/jenkins
 
 # Dependencies plus vim for convenience, wget if you're trying to get fancy
 RUN apt-get update \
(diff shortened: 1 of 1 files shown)


svanoort:
@olivergondza This is similar to the templating approach I used for the sudo-able build images -- (see https://github.com/jenkinsci/packaging/blob/master/docker/build-sudo-images.sh and remember the ARG instruction was not widely available yet), but I have found and tested a better way with #54.  

There are a couple advantages to that method: 

* Doesn't require rebuilding the image to get user support (allows use of a single image)
* Grants sudo within the container (helpful in specific cases)

If you agree and +1 that, we could merge and have a hardened approach.

docker/Dockerfile
```diff
@@ -1,6 +1,15 @@
 FROM ubuntu:14.04
 MAINTAINER svanoort <samvanoort@gmail.com>
 
+# User id of the build user
+ARG uid=4242
```
svanoort: Why that particular user?

docker/Dockerfile
```diff
@@ -1,6 +1,15 @@
...
+ENV uid $uid
+
+# Group id of the build user
+ARG gid=4242
+ENV gid $gid
+
+RUN groupadd builder -g $gid && useradd builder -u $uid -g $gid -m -d /home/builder
```
svanoort: Why 'builder' and not 'jenkins'?

docker/Dockerfile
olivergondza: I am trying to provide a default so existing automation will succeed to build the container. The limitation will still exist as before until updated.

olivergondza:
It is hard for me to comment here as I do not really understand why there are 2 kinds of containers - the sudoable and the build container. Aside from `ARG` instruction limited support (I was not aware of), I do not really like #54 any better as the main disadvantage of this path w.r.t. invocation from Pipeline (figuring out and passing all the ids to steps) is present in both. This one seems more straightforward to use to me, though.

docker/Dockerfile
svanoort: @olivergondza I think maybe jenkins & 1000 would make more sense for general users building jenkins packages, and then pass args for that specific builder?

svanoort:
@olivergondza The sudo-able approaches are pretty hacky -- I was trying to get something that worked but actually like the use of ARG here much better (and I think probably worth making a switchover to this in the next iteration of this codebase now that time has marched along).  But, the principle is the same.

I *think* this & #54 could actually be made to coexist, in all honesty - both could be merged together (but I do want both because I like being able to build my packages locally for testing changes, and this makes it easier).

docker/Dockerfile
olivergondza: I tried to avoid 1000 not to collide with non-system user accounts but common distributions' images do not seem to have any out of the box. So I am fine with that.

rtyler:
@svanoort From my perspective this can be merged, any reason not to?

svanoort:
@rtyler Yes, it is mergeable per previous

PR #68, opened 2016-09-14
[JENKINS-38206] Add the artifact name to the Windows installer

kwhetstone:
[JENKINS-38206](https://issues.jenkins-ci.org/browse/JENKINS-38206)

Normally the signed file is a random string of characters, which doesn't make sense.  By adding the /d (description ) here, the name of the requesting program will match the .msi file.

@reviewbybees

### DIFF
--- a/msi/build.sh
+++ b/msi/build.sh
@@ -51,7 +51,7 @@ candle -dJreDir="$JREDIR" -dWAR="$war" -dJavaExeId=$JavaExeId -nologo -ext WixUI
 light -o ${ARTIFACTNAME}.msi -sval -nologo -dcl:high -ext WixUIExtension -ext WixUtilExtension -ext WixFirewallExtension jenkins.wixobj jre.wixobj
 
 set +x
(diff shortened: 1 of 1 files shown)


armfergom:
@kwhetstone a ticket in to track this would be great :)

Other than that, this LGTM 🐝 . Have you tried it?

kwhetstone:
Hi Armando, yes, I've tried it on a Windows machine, and this produces a signature with "jenkins" as the app that will "make changes to your device".  I think this is fine since this is also the name of the msi, so it's at least consistent.

svanoort:
🐝  but probably needs a 1-pass by @jtnord

jtnord:
:bee: but I would prefer if you used the existing expression style. ${blah}

kwhetstone:
I changed it, could I get a rebee?  I don't have permission to merge, so I'll leave that to whoever does.

armfergom:
🐝

oleg-nenashev:
🐝

PR #70, opened 2016-09-26
[JENKINS-38134] Use same file for all shasums in the same folder

armfergom:
See [JENKINS-38134](https://issues.jenkins-ci.org/browse/JENKINS-38134)

Maintain just one SHA256SUMS file per folder with one entry per distributable in the folder instead of maintaining one SHA256SUMS file per distributable.

@reviewbybees @recena

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,9 +40,10 @@ docker.test: docker.images
 war: ${WAR}
 war.publish: ${WAR}
 	ssh ${SSH_OPTS} ${PKGSERVER} mkdir -p "'${WARDIR}/${VERSION}/'"
-	sha256sum ${WAR} > ${WAR_SHASUM}
+	rsync -avz -e "ssh ${SSH_OPTS}" "${WAR}" "${PKGSERVER}:${WARDIR}/${VERSION}/${ARTIFACTNAME}.war"
+	sha256sum ${WAR} | sed 's, .*/, ,' > ${WAR_SHASUM}
 	cat ${WAR_SHASUM}
-	rsync -avz -e "ssh ${SSH_OPTS}" "${WAR}" "${WAR_SHASUM}" "${PKGSERVER}:${WARDIR}/${VERSION}/${ARTIFACTNAME}.war"
+	cat ${WAR_SHASUM} | ssh ${SSH_OPTS} ${PKGSERVER} "cat >> ${WARDIR}/${VERSION}/SHA256SUMS"
 
 
 
--- a/msi/publish.sh
+++ b/msi/publish.sh
@@ -1,4 +1,5 @@
 #!/bin/bash -ex
-sha256sum ${MSI} > ${MSI_SHASUM}
+rsync -avz -e "ssh $SSH_OPTS" "${MSI}" "$PKGSERVER:$MSIDIR/"
+sha256sum ${MSI} | sed 's, .*/, ,' > ${MSI_SHASUM}
 cat ${MSI_SHASUM}
-rsync -avz -e "ssh $SSH_OPTS" "$MSI" "$MSI_SHASUM" "$PKGSERVER:$MSIDIR/"
+cat ${MSI_SHASUM} | ssh ${SSH_OPTS} ${PKGSERVER} "cat >> $MSIDIR/SHA256SUMS"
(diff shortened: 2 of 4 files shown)


recena:
The feedback from @daniel-beck and @olivergondza would be great.

LGTM.

daniel-beck:
It is unclear what the motivation here is. In general, I'd rather never again update a specific file's checksum file. Who knows whether we get it right the second time.

armfergom:
@daniel-beck the idea is to follow [this convention](http://cdimage.debian.org/debian-cd/8.6.0/amd64/iso-cd/)

Makefile
```diff
@@ -40,9 +40,10 @@ docker.test: docker.images
 war: ${WAR}
 war.publish: ${WAR}
 	ssh ${SSH_OPTS} ${PKGSERVER} mkdir -p "'${WARDIR}/${VERSION}/'"
-	sha256sum ${WAR} > ${WAR_SHASUM}
+	rsync -avz -e "ssh ${SSH_OPTS}" "${WAR}" "${PKGSERVER}:${WARDIR}/${VERSION}/${ARTIFACTNAME}.war"
+	sha256sum ${WAR} | sed 's, .*/, ,' > ${WAR_SHASUM}
```
svanoort: Is sha256 installed by the packaging dockerfile and called out in the dependencies for build (README and setup?)

svanoort: 🐛  for not generating the shasums before doing copy to server.  They can be rsynched with the WAR.

msi/publish.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash -ex
-sha256sum ${MSI} > ${MSI_SHASUM}
+rsync -avz -e "ssh $SSH_OPTS" "${MSI}" "$PKGSERVER:$MSIDIR/"
```
svanoort: 🐛  again for rsync before shasum

osx/publish.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash -ex
-sha256sum ${OSX} > ${OSX_SHASUM}
+rsync -avz -e "ssh $SSH_OPTS" "${OSX}" "$PKGSERVER:$OSXDIR/"
+sha256sum ${OSX} | sed 's, .*/, ,' > ${OSX_SHASUM}
```
svanoort: :bug: again for rsync before shasum

setup.mk
```diff
@@ -11,13 +10,15 @@ export TARGET:=target
 # jenkins-cli.jar
 export CLI:=${TARGET}/jenkins-cli.jar
 
+export WAR_SHASUM=${TARGET}/SHA256SUMS
+
 # where to generate MSI file?
 export MSI:=${TARGET}/msi/${ARTIFACTNAME}-${VERSION}.zip
-export MSI_SHASUM:=${MSI}-SHA256SUMS
+export MSI_SHASUM:=${TARGET}/msi/SHA256SUMS
```
svanoort: Excellent!

svanoort:
I can understand the motivation (sensible) but have entered bugs against the particular implementation used here.  Should be quick to resolve though.

daniel-beck:
:-1: This makes absolutely no sense for Jenkins. The Debian package files you reference all belong to the same release, published simultaneously, while you want append to a checksum file containing older releases' checksums.

You may not be aware of this, but we occasionally re-run packages for releases for some reason or other, and that will completely mess up your one file -- it'll have two entries for the same file name.

svanoort:
Concur with @daniel-beck on this, besides the bugs I've already logged. :-1:

Makefile
armfergom: It's not as easy as resynching as I wanted to follow [this convention](http://cdimage.debian.org/debian-cd/8.6.0/amd64/iso-cd/) as stated above.

recena:
🐝 and 👍

recena:
> but we occasionally re-run package creation for a release for some reason or other

If this can happen, obviously, this is not a good solution. A re-layout based on release instead of type of distributable.... 😄

armfergom:
Closing as this is superseded by #75

PR #71, opened 2016-09-27
[JENKINS-38437] Fix the debian package description

kwhetstone:
[JENKINS-38437](https://issues.jenkins-ci.org/browse/JENKINS-38437)

Sync the debian packaging message with the rhel and suse messages.  Since dbuild can't extract from a file like rpm, I had to pull it out into a variable in the `make` file.  Now the messaging is consistent across packages & will only have to updated in 1 place.  More information can be found in the JIRA ticket.

@reviewbybees


svanoort:
I did previous work around templating multi-line variables with the makefiles and it breaks in all sorts of complex and subtle ways.  🐛  

Hate to say it but we're far better off hardcoding.

For context: https://github.com/jenkinsci/packaging/pull/28

svanoort:
Looking closer at the issue, is there a reason we cannot do this by using the existing DESCRIPTION_FILE variable?   Not sure what you mean by "Since dbuild can't extract from a file like rpm"?

The file use *should* be entirely done withe branding script I did (replacing a brittle bash/perl combination that was posing problems), it just reads out to an environment variable and does substitutions:
https://github.com/jenkinsci/packaging/blob/master/bin/branding.py

kwhetstone:
The file won't work with direct since debian requires preceding spaces on all it's individual lines in the description, and a '.' on blank lines.  I've been playing with special string modification to get this to work

svanoort:
@kwhetstone What about creating a modified description file in place and building Debian separately using the modified version?  

Will play better with the issues that make has.

kwhetstone:
I hate having 2 different files simply because they have the same information just with spacing configurations.  Trying to preform os specific tasks doesn't really seem worth it in the grand scheme of things.  I added a `DESCRIPTION_FILE_DEB` that conforms to the Debian packaging requirements.  Pending review.

bin/branding-files.list
```diff
@@ -1 +1,2 @@
-DESCRIPTION_FILE
\ No newline at end of file
+DESCRIPTION_FILE
+DESCRIPTION_FILE_DEB
```
svanoort: 👎  on adding another variable to the branding list -- this already has entirely too many variables to set.  

Better to do transformation on the description file content  in the Debian build.sh (literally fold & sed commands to create a modified *version* of the file).  We can redefine the ENV variable to point at the modified file before calling the branding.py script, and then re-set it after.

**IMPORTANT** at no point do we want to try to read the file content into make variables.  They mangle it horribly.  File --> file transformation via a bash command is safer.

(Explained more about this in person).

deb/build/debian/control
```diff
@@ -11,6 +11,4 @@ Architecture: all
 Depends: ${misc:Depends}, daemon, adduser, procps, psmisc, net-tools, default-jre-headless (>= 2:1.7) | java7-runtime-headless
 Conflicts: hudson
 Replaces: hudson
-Description: continuous integration system
- Jenkins is an application that monitors executions of repeated jobs,
- such as building a software project or jobs run by cron.
+Description: @@DESCRIPTION_FILE_DEB@@
```
svanoort: Will become description_file (but with modified content).

svanoort: Alternately, you can define this environment variable within your build.sh script (assuming this is all done bash-side and doesn't go through the makefile madness it should work to read contents to/from variables).

kwhetstone:
Ok, I added a deb-specific temporary modification to the package description.  This adds the spacing and control-specific modifications.

recena:
@svanoort Could you take a look?

deb/build/build.sh
```diff
@@ -13,9 +13,15 @@ mkdir -p $D
 # so that multiple builds can go on concurrently
 cp -R $dir/* $D
 
+# Create a description temp file
+sed -i.bak -e 's/^\s*$/./' -e 's/^/ /' $DESCRIPTION_FILE
```
svanoort: YES.  This is exactly the approach I proposed -- just want to make sure the transformation is right: was this tested?  

I think you need the -r argument to use an extended regex.   Maybe best to provide the transformed output result in the description so we see that it generates correctly.

svanoort:
After seeing a sample output screenshot, looks good to me now.  Thanks for all the hard work @kwhetstone

svanoort:
@daniel-beck  / @olivergondza If you have no objections, will merge

daniel-beck:
Assuming this has been tested, fine with me.

Someone really should revise that description. I poked @rtyler to take a look at this.

deb/build/build.sh
```diff
@@ -13,9 +13,15 @@ mkdir -p $D
...
+
 # Expand variables in the definition
 "$BASE/bin/branding.py" $D/debian
 
+# Rewrite the file 
+mv "$DESCRIPTION_FILE.bak" "$DESCRIPTION_FILE"
```
samrocketman: What's the point of creating the description temp file and then overwriting it with the original (original created with `sed -i.bak...`)?

kwhetstone: Since the file is changed in place and we don't know the order of the processing the different packaging methods, we have to change the file back after processing the branding.

svanoort:
Merging as maintainerish, since it doesn't look like there are any merge blockers at this point.

deb/build/build.sh
samrocketman: I get it now; thanks.

PR #72, opened 2016-09-28
Ensure we're referring to the right base URL for packages

rtyler:
I /believe/ this is the right place to change things for INFRA-933, but I'm not
entirely sure whether this is the `RPM_URL` that will be used by `gen.rb` for
the generated page on https://pkg.jenkins.io/redhat

References [INFRA-933](https://issues.jenkins-ci.org/browse/INFRA-933)


recena:
👍

rtyler:
@daniel-beck or @olivergondza would you mind taking a look and merging if this is good enough? :)

PR #69, opened 2016-09-15
[JENKINS-38134] Generate shasum for war, windows and osx installers

armfergom:
[JENKINS-38134](https://issues.jenkins-ci.org/browse/JENKINS-38134)

@reviewbybees

PR #73, opened 2016-09-30
Rename checksum files to conventional FILENAME.sha256

daniel-beck:
Followup to #69 by @armfergom, renaming the checksum file.

CC @svanoort

### DIFF
--- a/setup.mk
+++ b/setup.mk
@@ -1,6 +1,6 @@
 # war file to release
 export WAR?=$(error Required variable WAR must point to the jenkins.war file you are packaging)
-export WAR_SHASUM=${WAR}-SHA256SUMS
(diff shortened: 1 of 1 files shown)


svanoort:
:+1:

recena:
👍 For the renaming but I prefer the idea here https://github.com/jenkinsci/packaging/pull/70, it should be reviewed though. Only one checksum file per folder.

armfergom:
👍 for the renaming, although I think this PR si no longer applicable since #69 was reverted. I'd like to create another PR to replace #69 and will take into account this naming proposal. Thanks @daniel-beck !

daniel-beck:
Superseded by #75

PR #74, opened 2016-10-02
Revert "[JENKINS-38134] Generate shasum for war, windows and osx inst…

daniel-beck:
…allers"

This reverts commit 916f556aad4d6fbbc7c8044c3056005cb29f652f.

Reverts #69 since [it seems to accidentally change the directory structure](https://github.com/jenkinsci/packaging/pull/69#discussion_r81468558).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,9 +40,7 @@ docker.test: docker.images
 war: ${WAR}
 war.publish: ${WAR}
 	ssh ${SSH_OPTS} ${PKGSERVER} mkdir -p "'${WARDIR}/${VERSION}/'"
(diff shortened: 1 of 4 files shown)


daniel-beck:
CC @armfergom FYI

oleg-nenashev:
👍 as a hotfix

daniel-beck:
I don't know when exactly @kohsuke returns to create the new weekly and LTS release, but I believe #69 would impact directory structure, so I'll just merge this now to be safe (in the hopes it doesn't break anything else).

armfergom:
@daniel-beck fair enough, thanks for this. I'l think of a different proposal now that I know more. Sorry for the trouble.

PR #75, opened 2016-10-04
[JENKINS-38134] Generate checksums for war, Windows and OS X distributables

armfergom:
[JENKINS-38134](https://issues.jenkins-ci.org/browse/JENKINS-38134), supersedes #69 and #70

Generate and publish shasums for WAR and OSX and Windows installers.

@daniel-beck @rtyler @olivergondza can I have your thoughts?

@reviewbybees

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,9 @@ setup:
 
 package: war msi osx deb rpm suse
 
-publish: war.publish msi.publish osx.publish deb.publish rpm.publish suse.publish
+publish: 
+	bash ./toolAvailability.sh
+	war.publish msi.publish osx.publish deb.publish rpm.publish suse.publish
 
 test: deb.test rpm.test suse.test
 
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,10 @@ docker.test: docker.images
 war: ${WAR}
 war.publish: ${WAR}
 	ssh ${SSH_OPTS} ${PKGSERVER} mkdir -p "'${WARDIR}/${VERSION}/'"
+	sha256sum ${WAR} | sed 's, .*/, ,' > ${WAR_SHASUM}
+	cat ${WAR_SHASUM}
 	rsync -avz -e "ssh ${SSH_OPTS}" "${WAR}" "${PKGSERVER}:${WARDIR}/${VERSION}/${ARTIFACTNAME}.war"
+	rsync -avz -e "ssh ${SSH_OPTS}" "${WAR_SHASUM}" "${PKGSERVER}:${WARDIR}/${VERSION}/"
 
 
 
--- a/msi/publish.sh
+++ b/msi/publish.sh
@@ -1,2 +1,4 @@
 #!/bin/bash -ex
-rsync -avz -e "ssh $SSH_OPTS" "$MSI" "$PKGSERVER:$MSIDIR/"
+sha256sum ${MSI} | sed 's, .*/, ,' > ${MSI_SHASUM}
+cat ${MSI_SHASUM}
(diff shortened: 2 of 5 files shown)


daniel-beck:
Curious, why not do the same for the Linux packages? Is it somehow not needed there?

armfergom:
@daniel-beck I believe the package repositories have their own verification mechanism. We'd need to ask ourselves if people manually download the linux packages and do not use the repos. If that is the case, we may also want to generate the shasums for those packages.

recena:
🐝

armfergom:
@daniel-beck may we move forward and merge? WDYT?

daniel-beck:
@olivergondza As one of the people running these scripts, any opinion?

recena:
@olivergondza any comment?

daniel-beck:
Do we know that `sha256sum` is on every system running these commands? If not, a `which sha256sum >/dev/null 2>&1 || { echo "sha256sum not found" >&2 ; exit 1 ; }` at a central location would be useful.

armfergom:
@daniel-beck do we want to abort the whole execution if that command is not found? Also, just a concern, if we are checking the existence of this tool at the beginning, shouldn't we doing the same for many other tools being used by these scripts? for example (but not only) `rsync`

daniel-beck:
@armfergom Good point, but we know that it works with `rsync` today, so that's less of a concern in practice (but should be added anyway). I don't know what kind of system runs these commands and how certain it is that `sha256sum` exists.

An additional (actual) problem is that these bash scripts are designed to fail early (`-e` is `errexit`) but won't fail when `sha256sum` isn't present, as `pipefail` isn't set. (And while we're at it, `nounset` would be nice, too.)

Other than that, there's been plenty of time for @olivergondza to review and we got no response, so I guess there's not going to be a review by him.

svanoort:
Concur with @daniel-beck.  Also I *swear* that I asked about the command missing in a previous related PR.  

Nounset can be somewhat problematic if there is already handling to provide defaults when not set in the scripts (some have it, we did it for the packaging tests).  Pipefail is a big one to include, I agree.

armfergom:
@daniel-beck @svanoort Added one more commit to check the availability of `sha256sum`. We can leave `pipefail` and/or `nounset` enablement for a follow up PR I think.

daniel-beck:
`pipefail` I would consider relevant (as in, it's a bug introduced in this change) as the scripts are configured to fail fast, but won't in the case of a `sha256sum` problem.

armfergom:
You got it @daniel-beck . Enabled `pipefail` where relevant.

daniel-beck:
@armfergom I have never seen this particular invocation, always having `set -o pipefail` in the script body. Have you tried to run this?

armfergom:
@daniel-beck I can do that too. Have tested that my approach works though.

Execution of:

```
#!/bin/bash -eu -o pipefail

ls | asdfasdf something | cat
echo "HI!!"
```

outputs:

```
./testScript.sh: line 3: asdfasdf: command not found
```

While execution of (note `-o pipefail` is not present):

```
#!/bin/bash -eu

ls | asdfasdf something | cat
echo "HI!!"
```

outputs:

```
./testScript.sh: line 3: asdfasdf: command not found
HI!!
```

daniel-beck:
In that case it's fine.

armfergom:
Great, so if there are no other concerns, it would be awesome if we could get this merge :)

recena:
@olivergondza @rtyler @svanoort ping?

olivergondza:
I have no opinion on this.

PR #79, opened 2016-11-01
Unblocking the release process

kohsuke:
I had to make these changes to unblock the release process from completion. It consists of two changes:

* Rolling back a change made by @armfergom which appears to be a genuine error
* Bypassing a strange bash error on `-o pipefail` in the shebang line

@reviewbybees

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,9 +24,7 @@ setup:
 
 package: war msi osx deb rpm suse
 
-publish: 
-	bash ./toolAvailability.sh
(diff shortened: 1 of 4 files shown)


kohsuke:
@abayer thinks https://gist.github.com/abayer/af7886a5172580b20e8164b882820764 is the correct fix for JENKINS-38134, and that looks good to me.

oleg-nenashev:
LGTM 🐝

svanoort:
🐝

recena:
@kohsuke I think @armfergom will back here so that we have `sha256sum`

PR #81, opened 2016-11-04
Remove the unnecessary .htaccess overlay contents for redirects

rtyler:
These are all now managed in the jenkins-infra/jenkins-infra Puppet repository.
As they have no per-version contents, it doesn't make sense for them to exist in
this repository anyways.


svanoort:
:+1:

PR #87, opened 2016-12-07
[JENKINS-23273] RPM post-install should not do a "chown -R ${JENKINS_USER:-jenkins} ${JENKINS_HOME:-/var/lib/jenkins}"

raul-arabaolaza:
[JENKINS-23273](https://issues.jenkins-ci.org/browse/JENKINS-23273)

Change ownership only if needed:

* Now `/var/cache/%{name}` `/var/log/%{name}` and workdir files are chowned only if parent folder is not already owned by appropriate user
* Also added sticky bit to that folders so only the appropriate user can create files there

@reviewbybees (specially, @kohsuke and @daniel-beck)

### DIFF
--- a/rpm/build/SPECS/jenkins.spec
+++ b/rpm/build/SPECS/jenkins.spec
@@ -76,22 +76,46 @@ rm -rf "%{buildroot}"
 /usr/sbin/useradd -g %{name} -s /bin/false -r -c "@@SUMMARY@@" \
 	-d "%{workdir}" %{name} &>/dev/null || :
 
+  # Used to decide later if we should perform a chown in case JENKINS_INSTALL_SKIP_CHOWN is false
+  # Check if a previous installation exists, if so use the configured JENKINS_USER to generate a files file for later use
+  # And check the JENKINS_HOME value and existing owners of work, log and cache dir, need to to this check
+  # here because the %files directive overwrites folder owners, I have not found a simple way to make the
+  # files directive to use JENKINS_USER as owner.
+  if [ -f "/etc/sysconfig/%{name}" ]; then
+      . /etc/sysconfig/%{name}
+      touch /tmp/cacheowner
+      ls -ld /var/cache/%{name} | awk 'NR==1 {print $3}' > /tmp/cacheowner
+      touch /tmp/logowner
+      ls -ld /var/log/%{name} | awk 'NR==1 {print $3}' >  /tmp/logowner
+      touch /tmp/workdirowner
+      ls -ld ${JENKINS_HOME:-%{workdir}}| awk 'NR==1 {print $3}'  > /tmp/workdirowner
+  fi
+
 %post
 /sbin/chkconfig --add %{name}
 
 # Ensure the right ownership on files only if not owned by JENKINS_USER
 . /etc/sysconfig/%{name}
 if test x"$JENKINS_INSTALL_SKIP_CHOWN" != "xtrue"; then
-    owner=$(ls -ld /var/cache/%{name} | awk 'NR==1 {print $3}')
-    if [ "$owner" != "${JENKINS_USER:-%{name}}" ] ; then
+    if [ -f "/tmp/cacheowner" ]; then
+      cacheOwner=$(cat /tmp/cacheowner)
+      rm -f /tmp/cacheowner
+    fi
+    if [ "${JENKINS_USER:-%{name}}" != "%{name}" ] || [ "${cacheOwner:-%{name}}" != "${JENKINS_USER:-%{name}}" ] ; then
         chown -R ${JENKINS_USER:-%{name}} /var/cache/%{name}
     fi
-    owner=$(ls -ld /var/log/%{name} | awk 'NR==1 {print $3}')
-    if [ "$owner" != "${JENKINS_USER:-%{name}}" ] ; then
+    if [ -f "/tmp/logowner" ]; then
+      logOwner=$(cat /tmp/logowner)
+      rm -f /tmp/logowner
+    fi
+    if [ "${JENKINS_USER:-%{name}}" != "%{name}" ] || [ "${logOwner:-%{name}}" != "${JENKINS_USER:-%{name}}" ] ; then
         chown -R ${JENKINS_USER:-%{name}} /var/log/%{name}
     fi
-    owner=$(ls -ld ${JENKINS_HOME:-%{workdir}}| awk 'NR==1 {print $3}')
-    if [ "$owner" != "${JENKINS_USER:-%{name}}" ] ; then
+    if [ -f "/tmp/workdirowner" ]; then
+      workdirOwner=$(cat /tmp/workdirowner)
+      rm -f /tmp/workdirowner
+    fi
+    if [ "${JENKINS_USER:-%{name}}" != "%{name}" ] || [ "${workdirOwner:-%{name}}" != "${JENKINS_USER:-%{name}}" ] ; then
         chown -R ${JENKINS_USER:-%{name}} ${JENKINS_HOME:-%{workdir}}
     fi
 fi
--- a/rpm/build/SPECS/jenkins.spec
+++ b/rpm/build/SPECS/jenkins.spec
@@ -79,42 +79,6 @@ rm -rf "%{buildroot}"
 %post
 /sbin/chkconfig --add %{name}
 
-# If we have an old hudson install, rename it to jenkins
-if test -d /var/lib/hudson; then
-    # leave a marker to indicate this came from Hudson.
-    # could be useful down the road
-    # This also ensures that the .??* wildcard matches something
-    touch /var/lib/hudson/.moving-hudson
-    # remove the hudson default update site to prevent it to be used by jenkins
-    if test -e /var/lib/hudson/updates/default.json; then
-      rm /var/lib/hudson/updates/default.json
-    fi
-    mv -f /var/lib/hudson/* /var/lib/hudson/.??* /var/lib/%{name}
-    rmdir /var/lib/hudson
-    find /var/lib/%{name} -user hudson -exec chown %{name} {} + || true
-fi
-if test -d /var/run/hudson; then
-    mv -f /var/run/hudson/* /var/run/%{name}
-    rmdir /var/run/hudson
-fi
-
-# Ensure the right ownership on files only if not owned by JENKINS_USER
-. /etc/sysconfig/%{name}
-if test x"$JENKINS_INSTALL_SKIP_CHOWN" != "xtrue"; then
-  owner=$(ls -ld /var/cache/%{name} | awk 'NR==1 {print $3}')
-  if [ "$owner" != "${JENKINS_USER:-%{name}}" ] ; then
-    chown -R ${JENKINS_USER:-%{name}} /var/cache/%{name}
-  fi
-  owner=$(ls -ld /var/log/%{name} | awk 'NR==1 {print $3}')
-  if [ "$owner" != "${JENKINS_USER:-%{name}}" ] ; then
-   chown -R ${JENKINS_USER:-%{name}} /var/log/%{name}
-  fi
-  owner=$(ls -ld ${JENKINS_HOME:-%{workdir}}| awk 'NR==1 {print $3}')
-  if [ "$owner" != "${JENKINS_USER:-%{name}}" ] ; then
-    chown -R ${JENKINS_USER:-%{name}} ${JENKINS_HOME:-%{workdir}}
-   fi
-fi
-
 %preun
 if [ "$1" = 0 ] ; then
     # if this is uninstallation as opposed to upgrade, delete the service
--- a/rpm/build/SPECS/jenkins.spec
+++ b/rpm/build/SPECS/jenkins.spec
@@ -93,7 +93,7 @@ rm -rf "%{buildroot}"
 %post
 /sbin/chkconfig --add %{name}
 
-function chownIfNeccesary {
+function chownIfNecessary {
   logger -t %{name}.installer "Checking ${2} ownership"
   if [ -f "${1}" ] ; then
     owner=$(cat $1)
@@ -118,9 +118,9 @@ function chownIfNeccesary {
 # the existing semantics
 . /etc/sysconfig/%{name}
 if test x"$JENKINS_INSTALL_SKIP_CHOWN" != "xtrue"; then
-      chownIfNeccesary "/tmp/%{name}.installer.cacheowner"  /var/cache/%{name}
-      chownIfNeccesary "/tmp/%{name}.installer.logowner"  /var/log/%{name}
-      chownIfNeccesary "/tmp/%{name}.installer.workdirowner"  ${JENKINS_HOME:-%{workdir}}
+      chownIfNecessary "/tmp/%{name}.installer.cacheowner"  /var/cache/%{name}
+      chownIfNecessary "/tmp/%{name}.installer.logowner"  /var/log/%{name}
+      chownIfNecessary "/tmp/%{name}.installer.workdirowner"  ${JENKINS_HOME:-%{workdir}}
 fi
 
 %preun
--- a/rpm/build/SPECS/jenkins.spec
+++ b/rpm/build/SPECS/jenkins.spec
@@ -77,8 +77,7 @@ rm -rf "%{buildroot}"
 	-d "%{workdir}" %{name} &>/dev/null || :
 
   # Used to decide later if we should perform a chown in case JENKINS_INSTALL_SKIP_CHOWN is false
-  # Check if a previous installation exists, if so use the configured JENKINS_USER to generate a files file for later use
-  # And check the JENKINS_HOME value and existing owners of work, log and cache dir, need to to this check
+  # Check if a previous installation exists, if so check the JENKINS_HOME value and existing owners of work, log and cache dir, need to to this check
   # here because the %files directive overwrites folder owners, I have not found a simple way to make the
   # files directive to use JENKINS_USER as owner.
   if [ -f "/etc/sysconfig/%{name}" ]; then
@@ -94,7 +93,10 @@ rm -rf "%{buildroot}"
 %post
 /sbin/chkconfig --add %{name}
 
-# Ensure the right ownership on files only if not owned by JENKINS_USER
+# Ensure the right ownership on files only if not owned by JENKINS_USER and JENKINS_USER
+# != %{name}, namely all cases but the default one (configured for %{name} owned by %{name})
+# In any case if JENKINS_INSTALL_SKIP_CHOWN is true we do not chown anything to maintain
+# the existing semantics
 . /etc/sysconfig/%{name}
 if test x"$JENKINS_INSTALL_SKIP_CHOWN" != "xtrue"; then
     if [ -f "/tmp/cacheowner" ]; then
--- a/rpm/build/SPECS/jenkins.spec
+++ b/rpm/build/SPECS/jenkins.spec
@@ -81,17 +81,36 @@ rm -rf "%{buildroot}"
   # here because the %files directive overwrites folder owners, I have not found a simple way to make the
   # files directive to use JENKINS_USER as owner.
   if [ -f "/etc/sysconfig/%{name}" ]; then
+      logger -t %{name}.installer "Found previous config file /etc/sysconfig/%{name}"
       . /etc/sysconfig/%{name}
-      touch /tmp/cacheowner
-      ls -ld /var/cache/%{name} | awk 'NR==1 {print $3}' > /tmp/cacheowner
-      touch /tmp/logowner
-      ls -ld /var/log/%{name} | awk 'NR==1 {print $3}' >  /tmp/logowner
-      touch /tmp/workdirowner
-      ls -ld ${JENKINS_HOME:-%{workdir}}| awk 'NR==1 {print $3}'  > /tmp/workdirowner
+      stat --format=%U /var/cache/%{name} > /tmp/%{name}.installer.cacheowner
+      stat --format=%U /var/log/%{name}  >  /tmp/%{name}.installer.logowner
+      stat --format=%U ${JENKINS_HOME:-%{workdir}}  > /tmp/%{name}.installer.workdirowner
+  else
+      logger -t %{name}.installer "No previous config file /etc/sysconfig/%{name} found"
   fi
 
 %post
 /sbin/chkconfig --add %{name}
+
+function chownIfNeccesary {
+  logger -t %{name}.installer "Checking ${2} ownership"
+  if [ -f "${1}" ] ; then
+    owner=$(cat $1)
+    rm -f $1
+    logger -t %{name}.installer "Found previous owner of ${2}: ${owner} "
+  fi
+  if  [ "${owner:-%{name}}" != "${JENKINS_USER:-%{name}}" ] ; then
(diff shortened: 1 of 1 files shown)


daniel-beck:
> Also added sticky bit to that folders so only the appropriate user can create files there

Uh… what?

> The restricted deletion flag or sticky bit is a single bit, whose interpretation depends on the file type. For directories, it prevents unprivileged users from removing or renaming a file in the directory unless they own the file or the directory; this is called the restricted deletion flag for the directory, and is commonly found on world-writable directories like /tmp.

How is that relevant for a folder that only a single user should have write access to anyway?

raul-arabaolaza:
@daniel-beck 

> How is that relevant for a folder that only a single user should have write access to anyway?

If you use the sticky bit you know that files created by jenkins can not be removed even by users with write and read permission, you need a root to remove the sticky bit before that or be the owner. Call that an extra layer of protection.

You believe I am being paranoid?

daniel-beck:
> You believe I am being paranoid?

Why are we not just restricting who gets to read JENKINS_HOME? Any process that involves direct JENKINS_HOME file access from another user seems pretty broken to me. A 'worst practice' we should not support.

Anyway, unrelated what this PR is supposed to do.

recena:
My suggestion is to remove this part of the script. Do you think that an upgrade from Hudson will work fine? IMHO, it does not make sense.

raul-arabaolaza:
@daniel-beck Not unrelated IMO, because before this change a `chown` were always executed when no `JENKINS_INSTALL_SKIP_CHOWN=false` property were set and after this change that `chown` would be executed conditionally *based on parent folder ownership*, so adding the setgid bit to ensure proper ownership of files on that folder seems relevant to me.

> When setgid permission is applied to a directory, files that were created in this directory belong to the group to which the directory belongs, not the group to which the creating process belongs. Any user who has write and execute permissions in the directory can create a file there. However, the file belongs to the group that owns the directory, not to the user's group ownership.

raul-arabaolaza:
I believe I will just wait for #88 to finish the smoke tests, is more than probable that those tests give us more insights about the need to migrate previously existing hudson installations and so we may decide if we need to keep this code or we can remove it

raul-arabaolaza:
I have performed some smoke tests for Hudson -> Jenkins migration

* Installed `Hudson 3.3.1` via rpm on a CentOS machine, created a freestyle job and executed it
* Removed without purge the hudson package and installed via rpm `Jenkins 2.32`

* Installation looks fine  
* jenkins service starts normally 
* The created job appears in the jenkins dashboard and executes fine

**BUT There are several warnings in the UI**

  * One warning related to hudson plugins installed not being compatible with jenkins 
```
There are dependency errors loading some plugins:
XPath Provider Plugin v1.0.3
     You must update Jenkins from v2.32 to v3.0.0 or later to run this plugin.
JNA Native Support Plugin v3.0.4
     You must update Jenkins from v2.32 to v3.0.0 or later to run this plugin.
```
  * One warning relative to old data format due to incompatibilities between hudson job xml and the Jenkins one
   * **A warning saying that there is a new version of jenkins available where the `download` links points to a hudson war**
```
New version of Jenkins (3.3.3) is available for download (changelog). 
```  
And other warnings that are expected like using an unsecured instance and so on. Please see the attached screenshoot
migrate

The problem relative to a new version of jenkins available pointing to a hudson war is solved by using the `Check now` button in the `Advanced` tab of the `Manage Plugins` section

The problem with the incompatible plugins is solved by manually removing them from the `plugins` folder in `JENKINS_HOME` 

The old data problem can also be solved using the standard process for manage old data, re saving the file without the problematic fields and dismissing the message

So in general this simple smoke shows that migration from `hudson` to `jenkins` is possible although (as expected) not automatic for simple cases. For more complex scenarios more manual tweaking would be needed to solve conflicts, but that is expected.

samrocketman:
> A warning saying that there is a new version of jenkins available where the download links points to a hudson war

Perhaps an upgrade should involve deleting `$JENKINS_HOME/updates/*.json`. In my Jenkins config, there's a `default.json` for the default update site where it caches the latest versions of Jenkins and plugins from the update site.

raul-arabaolaza:
@samrocketman You are right, I have incorporated your suggestion

daniel-beck:
We have never supported a "cross-grade" from Hudson 2.x+ to Jenkins. There's going to be so much broken due to incompatibility I wouldn't even consider supporting even the easy bits.

raul-arabaolaza:
@daniel-beck So you agree with @recena suggestion on remove the part of the script related to that? If we do not support migration there is no need for the chown and instead of adding guards and ifs i can simply delete it.

daniel-beck:
@raul-arabaolaza 

> If we do not support migration there is no need for the chown

We support "upgrade" from Hudson 1.396 or lower to Jenkins. `chown` makes sense then. But anything related to more recent Hudson releases, we don't support -- and AFAIU this includes the very weird update site metadata.

raul-arabaolaza:
@daniel-beck And does it make sense to support an upgrade only from a version that is 6 years old? I mean, does it make sense today to include the code to suport that? I can understand the support at the time of the jenkins/hudson fork and for some time after that. But the current hudson version is 3.3.3. I feel very difficult to imagine anyone with a hudson 1.396 or lower needing/willing to migrate to jenkins today...

recena:
👍  for removing this part the code.

raul-arabaolaza:
Ok, so I have implemented my proposal on this PR, I believe it does not make sense anymore to support the hudson -> jenkins migration and so I have deleted the migration code. As a consecuence IMO the chown operation does not make sense anymore and so I have also removed it

daniel-beck:
@raul-arabaolaza So what you're saying is that you would like to reopen [JENKINS-12231](https://issues.jenkins-ci.org/browse/JENKINS-12231)?

aheritier:
:bee:
I prefer to do nothing (like proposed here) than doing something wrong 99% of times (useless chown) for something used by 0.00001% (migration from hudson to jenkins)

recena:
🐝  and 👍  for removing this part the code.

raul-arabaolaza:
@daniel-beck Would you agree to this PR if i apply the chown conditionally only in the case of the owner of the involved folders is not JENKINS_USER?. That is, not by default and or controlled by an environment variable? And that I remove the hudson -> jenkins migration code?

rsandell:
So why was the hudson -> jenkins migration removed???

raul-arabaolaza:
@rsandell Because the migration was supported only for Hudson 1.396 or lower (as stated before in the comments), that is a 6+ years old version of Hudson (currently in 3.3.3 version). IMHO is unrealistic to expect Hudson 1.396 (or lower) users  willing to move to Jenkins today. Obviously I would have a different opinion if the support where intended for recent versions of Hudson, but that is not the case.

recena:
@daniel-beck 

> Appears to break Jenkins updates for users with custom JENKINS_USER.

For those scenarios (corner cases) IMO, we could add a specific entry in the documentation.

I don't know why, but we need to go forward with this JIRA issue.

raul-arabaolaza:
I have talked with @daniel-beck and so I have implemented the changes that I believe he would find reasonable

raul-arabaolaza:
So, I have tested this scenarios:

1. An existing instance`s folders are owned by jenkins and configured for jenkins user: In this case no _chown_ is needed. With this changes no _chown_ is executed and the instance is running after upgrade
2. An existing instance`s folders are owned by a custom user and the instance is configured for a jenkins user: In this case a _chown_ is needed to ensure the instance works after upgrade. With this changes the _chown_ is executed and the instance is running after upgrade
3. An existing instance`s folders are owned by default jenkins user and the instance is configured for a custom user: In this case a _chown_ is needed to ensure the instance works after upgrade. With this changes the _chown_ is executed and the instance is running after upgrade
4. An existing instance`s folders is owned by a custom user and the instance is configured for that custom user: In this case a _chown_ is needed to ensure the instance works after upgrade. Due to new files being owned by default jenkins user (the %files directive in the rpm is to blame for this) With this changes the _chown_ is executed and the instance is running after upgrade

The initial intention was to skip _chown_ also in case four, but that results in a non runable instance (due to new files being copied being owned by the default jenkins user). I try to change the installer to make sure new files are owned by the configured user and so skip the _chown_, but I was not able to find an easy way. So instead of overly complicate the installer I opted to skip _chown_ only in the default case, this IMHO results in [JENKINS-23273](https://issues.jenkins-ci.org/browse/JENKINS-23273) being addressed whilst [JENKINS-12231](https://issues.jenkins-ci.org/browse/JENKINS-12231) has not to be reopen.

Please note that _JENKINS_INSTALL_SKIP_CHOWN_ semantics are preserved so a _true_ value results in no _chown_ executed in any case

@daniel-beck WDYT?

daniel-beck:
@raul-arabaolaza 

> the %files directive in the rpm is to blame for this

This is exactly the case I was worried about but couldn't find docs on. Glad we tested this.

AFAIU we don't need a recursive chown, but just a simple chown of a few specific files in this case?

raul-arabaolaza:
@daniel-beck AFAIK there is no way to know in advance which are the files I need to chown, I can do a recursive test to locate the files that need chowning but IMHO that would complicate the code and I am not sure if the test itself and the individual chowns are going to be quicker than the recursive chown.

raul-arabaolaza:
I believe this is not perfect, but fair enough, we can always create more tickets and PR if the need to further improve this arises

samrocketman:
You can use the `find` command to search for files needing to be chown and only chown those (directories are also files in Linux).  e.g.

```
find /path \( ! -user jenkins -o ! -group jenkins \) -execdir chown jenkins: {} +
```

daniel-beck:
@raul-arabaolaza 

> AFAIK there is no way to know in advance which are the files I need to chown

Assuming the user takes over the responsibility to `chown` as they change the account Jenkins runs as, this just leaves the explicitly mentioned files from `%files` the installer needs to take care of in case 4, right?

raul-arabaolaza:
@daniel-beck You are right, in scenario 4 I only need to chown the folders no to do a recursive chown, with this change the result is:

1. An existing instance's folders are owned by `jenkins` and configured for `jenkins` user: In this case no _chown_ is needed. With this changes no _chown_ is executed and the instance is running after upgrade
2. An existing instance's folders are owned by a custom user and the instance is configured for a `jenkins` user: In this case a recursive _chown_ is needed to ensure the instance works after upgrade. With this changes the recursive _chown_ is executed and the instance is running after upgrade. Folders after upgrade are owned by `jenkins` user (this is the current behaviour)
3. An existing instance's folders are owned by `jenkins` user and the instance is configured for a custom user: In this case a recursive _chown_ is needed to ensure the instance works after upgrade. With this changes the recursive _chown_ is executed and the instance is running after upgrade. Folders after upgrade are owned by the custom user (this is the current behaviour)
4. An existing instance's folders is owned by a custom user and the instance is configured for that custom user: In this case a _chown_ for the folders is needed to ensure the instance works after upgrade. With this changes the _chown_ is executed for the folders and the instance is running after upgrade. Folders after upgrade are owned by custom user

rpm/build/SPECS/jenkins.spec
```diff
@@ -76,30 +76,57 @@ rm -rf "%{buildroot}"
...
 	-d "%{workdir}" %{name} &>/dev/null || :
 
+  # Used to decide later if we should perform a chown in case JENKINS_INSTALL_SKIP_CHOWN is false
+  # Check if a previous installation exists, if so check the JENKINS_HOME value and existing owners of work, log and cache dir, need to to this check
+  # here because the %files directive overwrites folder owners, I have not found a simple way to make the
+  # files directive to use JENKINS_USER as owner.
+  if [ -f "/etc/sysconfig/%{name}" ]; then
+      . /etc/sysconfig/%{name}
+      touch /tmp/cacheowner
+      ls -ld /var/cache/%{name} | awk 'NR==1 {print $3}' > /tmp/cacheowner
```
daniel-beck: This seems unnecessarily brittle. Why not `stat --format=%U`?

rpm/build/SPECS/jenkins.spec
```diff
@@ -76,30 +76,57 @@ rm -rf "%{buildroot}"
 /usr/sbin/useradd -g %{name} -s /bin/false -r -c "@@SUMMARY@@" \
 	-d "%{workdir}" %{name} &>/dev/null || :
 
+  # Used to decide later if we should perform a chown in case JENKINS_INSTALL_SKIP_CHOWN is false
+  # Check if a previous installation exists, if so check the JENKINS_HOME value and existing owners of work, log and cache dir, need to to this check
+  # here because the %files directive overwrites folder owners, I have not found a simple way to make the
+  # files directive to use JENKINS_USER as owner.
+  if [ -f "/etc/sysconfig/%{name}" ]; then
+      . /etc/sysconfig/%{name}
+      touch /tmp/cacheowner
```
daniel-beck: No need to touch if you write to it in the following line.

Also, files should be named such that it's trivial to understand for admins who/what created them, e.g. if the installer fails to finish and the files are left behind. Also, file name collisions: Maybe append the current process ID (`$$`) or similar to prevent that? (Not sure if that works for installer scripts though)

rpm/build/SPECS/jenkins.spec
```diff
@@ -76,30 +76,57 @@ rm -rf "%{buildroot}"
...
+# Ensure the right ownership on files only if not owned by JENKINS_USER and JENKINS_USER
+# != %{name}, namely all cases but the default one (configured for %{name} owned by %{name})
+# In any case if JENKINS_INSTALL_SKIP_CHOWN is true we do not chown anything to maintain
+# the existing semantics
 . /etc/sysconfig/%{name}
 if test x"$JENKINS_INSTALL_SKIP_CHOWN" != "xtrue"; then
-   chown -R ${JENKINS_USER:-%{name}} /var/cache/%{name}
-   chown -R ${JENKINS_USER:-%{name}} /var/log/%{name}
-   chown -R ${JENKINS_USER:-%{name}} ${JENKINS_HOME:-%{workdir}}
+    if [ -f "/tmp/cacheowner" ]; then
```
daniel-beck: Seems to be a shell function may be in order here given the excessive copy/pasting.

rpm/build/SPECS/jenkins.spec
```diff
@@ -76,30 +76,57 @@ rm -rf "%{buildroot}"
...
+      cacheOwner=$(cat /tmp/cacheowner)
+      rm -f /tmp/cacheowner
+    fi
+    if  [ "${cacheOwner:-%{name}}" != "${JENKINS_USER:-%{name}}" ] ; then
+        chown -R ${JENKINS_USER:-%{name}} /var/cache/%{name}
```
daniel-beck: Is there a reasonable outlet for log output during this phase? If so, probably not a bad idea to do so. If not, also fine.

rpm/build/SPECS/jenkins.spec
```diff
@@ -76,30 +76,57 @@ rm -rf "%{buildroot}"
...
+      ls -ld /var/log/%{name} | awk 'NR==1 {print $3}' >  /tmp/logowner
+      touch /tmp/workdirowner
+      ls -ld ${JENKINS_HOME:-%{workdir}}| awk 'NR==1 {print $3}'  > /tmp/workdirowner
+  fi
+
 %post
 /sbin/chkconfig --add %{name}
 
-# If we have an old hudson install, rename it to jenkins
-if test -d /var/lib/hudson; then
```
daniel-beck: it's not clear to me why this gets rid of the Hudson compatibility code. AFAIU this change is completely unrelated. I'd understand if this would allow implementing this PR, but that doesn't seem to be the case?

raul-arabaolaza:
@daniel-beck 

> No need to touch if you write to it in the following line.
Also, files should be named such that it's trivial to understand for admins who/what created them, e.g. if  the installer fails to finish and the files are left behind. 

Changed, now file names are more descriptive

> Also, file name collisions: Maybe append the current process ID ($$) or similar to prevent that? (Not  sure if that works for installer scripts though)

No need for that because I regenerate the contents of the temp files on every installation, so even if the file is not removed and the next installation tries to use it the contents will be up to date 

> This seems unnecessarily brittle. Why not stat --format=%U?

No technical reason AFAIK, is just I am most used to `ls`, no strong opinion on my part here so I have changed it 

> Seems to be a shell function may be in order here given the excessive copy/pasting.

Yep, added it

> Is there a reasonable outlet for log output during this phase? If so, probably not a bad idea to do so. If not, also fine.

I have added it to the default sys log with the jenkins.installer tag, is the same place where the jenkins service is logging so I believe is coherent 

> it's not clear to me why this gets rid of the Hudson compatibility code. AFAIU this change is completely unrelated. I'd understand if this would allow implementing this PR, but that doesn't seem to be the case?

I am not sure if the doubt with this change is only because it seems unrelated or because you do not think is necessary. So I am going to try to answer both questions:

IMHO is necessary to remove this code because:
* The hudson -> jenkins migration was originally intended for a version that is six years old, so I believe (and you will need to show me a real recent case to convince me otherwise) is no more relevant, letting this code stay without a real need makes no sense. Ask yourself this question, if the installer were without this code and today someone makes a PR to include it Would you approve it or would you discard it as unnecessary? 
* The installer already declares itself as incompatible with hudson, so nobody should complain if no migration is executed 
* In case of need the migration can be easily done manually
About it being unrelated
Given that this code (IMHO) is no necessary and that I am doing a "big" change in the installer code and behaviour I truly believe is a good moment to get rid of it. Keeping the migration code would only mean that the end result of this PR would be not as good as it could be and that we are keeping unnecessary code. Just taking the clean up out to another, dependant PR is IMHO a waste of time. The diff is very small and if needed adding it back (I hope not) is very easy, so for my point of view there is no real reason to make another PR for that removal

daniel-beck:
@raul-arabaolaza 

> No need for that because I regenerate the contents of the temp files on every installation, so even if the file is not removed and the next installation tries to use it the contents will be up to date

This looks like a potential local attack vector: The installer is typically run as root, and it does not require that these files are owned by root (and if not run as root, writing to the files may well fail, and I don't know whether the script aborts in that case). An attacker can  create files with well known names in `/tmp`, write their own user name in there (in the root install case after the installer did), and have the installer grant them permissions to all of JENKINS_HOME.

> Ask yourself this question, if the installer were without this code and today someone makes a PR to include it Would you approve it or would you discard it as unnecessary?

I consider this to be a substantially different situation.

This mixes up two unrelated changes, which is rarely a good idea.

Still, I don't care enough about this -- so if you really want to remove it, fine with me.

samrocketman:
> No need for that because I regenerate the contents of the temp files on every installation

Why not use `mktemp` or `mktemp -d`?

raul-arabaolaza:
@daniel-beck 

> An attacker can create files with well known names in /tmp, write their own user name in there (in the root install case after the installer did), and have the installer grant them permissions to all of JENKINS_HOME.

The installer is using the contents of that files only to decide if a chown is needed, but the ownership of the `JENKINS_HOME` is granted to the `JENKINS_USER` defined in `/etc/sysconfig/jenkins` file. So in the case that an attacker is able to write the files before the first installation of jenkins in a machine the only result would be an unneeded `chown` but the final ownership of `JENKINS_HOME` will be `jenkins`

daniel-beck:
@samrocketman 

> Why not use `mktemp` or `mktemp -d`?

AFAIU this runs in different processes (`%pre` checks, `%post` acts), so no way to pass the file name around. Or is there?

> So in the case that an attacker is able to write the files before the first installation of jenkins in a machine the only result would be an unneeded `chown`

Good point. This seems to be acceptable.

raul-arabaolaza:
@samrocketman Is exactly as @daniel-beck said, thats also the reason I need to use files to store the owners

raul-arabaolaza:
@reviewbybees done

I believe this is ready for merge

dariver:
👍

aheritier:
+1000 to merge it @raul-arabaolaza  !!!

raul-arabaolaza:
Thanks @aheritier still waiting for a commiter to merge. I still have hope :)

aheritier:
@daniel-beck let's merge ? this is really a boring bug when an upgrade takes hours because we are doing a useless chown ...

daniel-beck:
@aheritier I don't maintain this component and won't merge a change I haven't tested. AFAICT @svanoort is the closest we have to a maintainer of this.

svanoort:
@daniel-beck I will try to take another look today, although before considering a change like this merge-worthy will have to do some testing.

raul-arabaolaza:
@svanoort Please ping me if you have any comment/bug/doubt

benh57:
Strong vote for this, incredibly frustrating. Happy to help test in any way, this bug/PR has been batting around a while.

recena:
Who could merge this?

benh57:
@svanoort The user side pain on this is high. With a large Jenkins installation, it can take many hours to update Jenkins. If you could take a look at it we'd appreciate it.

rpm/build/SPECS/jenkins.spec
```diff
@@ -76,30 +76,51 @@ rm -rf "%{buildroot}"
...
-if test -d /var/run/hudson; then
-    mv -f /var/run/hudson/* /var/run/%{name}
-    rmdir /var/run/hudson
-fi
-
-# Ensure the right ownership on files
+function chownIfNecessary {
+  logger -t %{name}.installer "Checking ${2} ownership"
+  if [ -f "${1}" ] ; then
+    owner=$(cat $1)
```
svanoort: 🐛  Unsafe if $1 includes spaces.

rpm/build/SPECS/jenkins.spec
```diff
@@ -76,30 +76,51 @@ rm -rf "%{buildroot}"
...
+    rm -f $1
```
svanoort: Again, unsafe if $1 includes spaces

rpm/build/SPECS/jenkins.spec
```diff
@@ -76,30 +76,51 @@ rm -rf "%{buildroot}"
...
+    logger -t %{name}.installer "Found previous owner of ${2}: ${owner} "
+  fi
+  if  [ "${owner:-%{name}}" != "${JENKINS_USER:-%{name}}" ] ; then
+    logger -t %{name}.installer "Previous owner of ${2} is different than configured JENKINS_USER... Doing a recursive chown of ${2} "
+    chown -R ${JENKINS_USER:-%{name}} $2
```
svanoort: Again, spaces.

rpm/build/SPECS/jenkins.spec
```diff
@@ -76,30 +76,51 @@ rm -rf "%{buildroot}"
...
+  elif [ "${JENKINS_USER:-%{name}}" != "%{name}" ] ; then
+    # User has changed ownership of files and JENKINS_USER, chown only the folder
+    logger -t %{name}.installer "Configured JENKINS_USER is different than default... Doing a non recursive chown of ${2} "
+    chown ${JENKINS_USER:-%{name}} $2
```
svanoort: Not listing spaces again after this, but fix all the variable uses which will break with spaces please.

raul-arabaolaza:
@svanoort I believe I have addressed the space issues by quoting the variables

recena:
@daniel-beck What do you think?

daniel-beck:
Quoting in the last looks reasonable, don't have the time to check whether it does everywhere it's needed, or the setup to do test runs.

svanoort:
As long as you concur the quoting is reasonable, I'm comfortable merging -- it sounds like @raul-arabaolaza has done quite extensive testing.

recena:
@daniel-beck @svanoort Thanks so much

PR #89, opened 2016-12-08
Fix the sed command invalid filename

hayderimran7:
.orig doesnt exist so should be removed in sed command.
```
sed -i .orig -e "s#@@MYUSERID@@#`id -u`#g" sudo-*/Dockerfile
sed: can't read .orig: No such file or directory
```

### DIFF
--- a/docker/build-sudo-images.sh
+++ b/docker/build-sudo-images.sh
@@ -4,7 +4,7 @@
 # Note: this modifies the original Dockerfiles so the current user can sudo
 
 cd "$(dirname "$0")"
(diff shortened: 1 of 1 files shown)


daniel-beck:
This script appears to use OS X (BSD) `sed` arguments. Not sure whether it's deliberate or not, but it's possible it's correct as is.

@svanoort Could you clarify?

svanoort:
@hayderimran7 This is indeed a result of platform-specific arguments due to BSD vs. Linux.  Any fix has to play nicely with both, and *if* I recall right, this one would complain but run in both cases.

PR #90, opened 2017-01-30
[JENKINS-41413] Working around https://github.com/jenkinsci/distfork-plugin/pull/6

kohsuke:
Make sure a failed windows installer build will not result in 0-byte msi file created at the expected location.


oleg-nenashev:
👍 . Ideally there should be an artifact size check at the end of the script. Just in case

samrocketman:
Perhaps a plugin option to fail on zero-size artifacts?

PR #91, opened 2017-03-08
Update descriptions

oleg-nenashev:
- [x] - No more "Continuous Integration Server"
- [x] - Update the common and MacOS descriptions, the original one is something ancient (external job monitor, etc.)

CC @jenkinsci/code-reviewers, esp. @rtyler and @daniel-beck . The description is IMHO not ideal, so I would appreciate any proposals.

### DIFF
--- a/branding/description-file
+++ b/branding/description-file
@@ -1,13 +1,5 @@
-Jenkins monitors executions of repeated jobs, such as building a software
-project or jobs run by cron. Among those things, current Jenkins focuses on the
-following two jobs:
-- Building/testing software projects continuously, just like CruiseControl or
-  DamageControl. In a nutshell, Jenkins provides an easy-to-use so-called
-  continuous integration system, making it easier for developers to integrate
-  changes to the project, and making it easier for users to obtain a fresh
-  build. The automated, continuous build increases the productivity.
-- Monitoring executions of externally-run jobs, such as cron jobs and procmail
-  jobs, even those that are run on a remote machine. For example, with cron,
-  all you receive is regular e-mails that capture the output, and it is up to
-  you to look at them diligently and notice when it broke. Jenkins keeps those
-  outputs and makes it easy for you to notice when something is wrong.
+Jenkins is an open source automation server which enables developers around the world to reliably automate  their development lifecycle processes of all kinds, including build, document, test, package, stage, deployment, static analysis and many more.
+
+Jenkins is being widely used in areas of Continuos Integration, Continuous Delivery, DevOps, and other areas. And it is not only about software, the same automation techniques can be applied in other areas like Hardware Engineering, Embedded Systems, BioTech, etc. 
+
+For information see https://jenkins.io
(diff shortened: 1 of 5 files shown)


branding/description-file
```diff
@@ -1,13 +1,5 @@
...
-  continuous integration system, making it easier for developers to integrate
-  changes to the project, and making it easier for users to obtain a fresh
-  build. The automated, continuous build increases the productivity.
-- Monitoring executions of externally-run jobs, such as cron jobs and procmail
-  jobs, even those that are run on a remote machine. For example, with cron,
-  all you receive is regular e-mails that capture the output, and it is up to
-  you to look at them diligently and notice when it broke. Jenkins keeps those
-  outputs and makes it easy for you to notice when something is wrong.
\ No newline at end of file
+Jenkins is an open source automation server which enables developers around the world to reliably automate  their development lifecycle processes of all kinds, including build, document, test, package, stage, deployment, static analysis and many more.
```
olivergondza: Double space: 'automate__their'.

daniel-beck:
> The description is IMHO not ideal, so I would appreciate any proposals.

Go with the blurb from https://jenkins.io/press/ ?

oleg-nenashev:
@daniel-beck I have started from that, but IMHO info like "In 2015, Jenkins surpassed 100,000 known installations making it the most widely deployed automation server." and "Jenkins, originally founded in 2006 as "Hudson", is one of the leading automation servers available. " does not make much sense for the description

branding/description-file
```diff
@@ -1,13 +1,5 @@
...
+
+Jenkins is being widely used in areas of Continuos Integration, Continuous Delivery, DevOps, and other areas. And it is not only about software, the same automation techniques can be applied in other areas like Hardware Engineering, Embedded Systems, BioTech, etc.
```
svanoort: I know why someone wants to mention "hardware engineering" etc, but IMO better to leave that last sentence off.  Description is intended to convey what it is and what it does, not be a catchall for all the things it *can* do, because those are too many to name.

osx/docs/JenkinsWelcomePanel.rtf
```diff
@@ -1,19 +1,15 @@
-{\rtf1\ansi\ansicpg1252\cocoartf1038\cocoasubrtf350
-{\fonttbl\f0\fswiss\fcharset0 ArialMT;}
-{\colortbl;\red255\green255\blue255;}
-{\*\listtable{\list\listtemplateid1\listhybrid{\listlevel\levelnfc23\levelnfcn23\leveljc0\leveljcn0\levelfollow0\levelstartat1\levelspace360\levelindent0{\*\levelmarker \{disc\}}{\leveltext\leveltemplateid1\'01\uc0\u8226 ;}{\levelnumbers;}\fi-360\li720\lin720 }{\listname ;}\listid1}}
-{\*\listoverridetable{\listoverride\listid1\listoverridecount0\ls1}}
+{\rtf1\ansi\ansicpg1252\cocoartf1504\cocoasubrtf810
```
svanoort: What's with all the markup here, @oleg-nenashev ?   The output when I view it seems to be more or less plaintext in appearance, is all this strictly needed?

svanoort:
👍  In general -- though I'd obviously prefer if comments are addressed.

osx/docs/JenkinsWelcomePanel.rtf
oleg-nenashev: It is RTF

PR #93, opened 2017-04-03
Add jenkins.service for systemd which is in use in debian

paladox:
This file is from @hashar @wikimedia See https://github.com/wikimedia/puppet/commit/07c8bb1ab32d9c4570ae2cb80761f2ce8d64302f

### DIFF
--- a/deb/build/debian/control
+++ b/deb/build/debian/control
@@ -2,7 +2,7 @@ Source: @@ARTIFACTNAME@@
 Section: devel
 Priority: extra
 Maintainer: @@AUTHOR@@
-Build-Depends: debhelper (>= 9)
+Build-Depends: debhelper (>= 9), dh-systemd
 Standards-Version: 3.9.4
 Homepage: @@HOMEPAGE@@
 
--- a/deb/build/debian/jenkins.service
+++ b/deb/build/debian/jenkins.service
@@ -1,0 +1,24 @@
+[Unit]
+Description=Jenkins Continuous Integration Server
+After=network.target
+
+[Service]
+Type=simple
+User=jenkins
+Group=jenkins
+SyslogIdentifier=jenkins
+UMask='0002'
+LimitNOFILE='8192'
+
+Environment=JENKINS_HOME=/var/lib/jenkins
+ExecStart=/usr/bin/java \
+    -jar /usr/share/jenkins/jenkins.war \
+    --webroot=/var/cache/jenkins/war \
+    --pluginroot=/var/cache/jenkins/plugins \
+    --httpPort='8080' \
+
+# We send SIGTERM(15), java exit code is 143 = 128 + 15
+SuccessExitStatus=143
+
+[Install]
+WantedBy=multi-user.target


svanoort:
@jtnord Is there any chance you could take a peek at this?  I know you've been talking about doing a SystemD service, and are more familiar with the format than me.

hashar:
It should probably source an environment file that contains the setting then the service would use an `EnvironmentFile`.

A couple gotchas at least:

There is no reload command defined. systemd defaults to sending SIGHUP but that is not handled by Jenkins


Jenkins logs INFO+ to stdout. That is captured by systemd which by default send it to the journal and thus systemd. That is quite spammy in journalctl and daemon.log.

I am solving that by creating a logging.properties file:
```
handlers=java.util.logging.ConsoleHandler
java.util.logging.SimpleFormatter.format = %4$s: [%2$s] %5$s%n
java.util.logging.ConsoleHandler.level = WARNING
java.util.logging.ConsoleHandler.formatter = java.util.logging.SimpleFormatter
```

And starting Jenkins with `-Djava.util.logging.config.file=/etc/jenkins/logging.properties`

Reference: https://gerrit.wikimedia.org/r/#/c/347877/


I don't think anything above is copyright worthy. If you rather be safe, feel free to attribute it to both authors (I have a joint copyright agreement with Wikimedia):
```
Antoine "hashar" Musso
Wikimedia Foundation Inc.
```

fuero:
I'm using this on CentOS 7. Running the SysV script incorrectly assigns the SELinux label initrc_t instead of unconfined_service_t. This is fixed by using a systemd initscript.

jenkins.service
```
[Unit]
Description=Jenkins build server

[Service]
Type=forking
User=jenkins
Group=jenkins
EnvironmentFile=-/etc/sysconfig/jenkins
ExecStart=/usr/local/bin/jenkins

[Install]
WantedBy=multi-user.target
```

/usr/local/bin/jenkins
```
#!/bin/sh
# Check for missing binaries (stale symlinks should not happen)                                                                                                                                                                                                                                                       
JENKINS_WAR="/usr/lib/jenkins/jenkins.war"
test -r "$JENKINS_WAR" || { echo "$JENKINS_WAR not installed";
        if [ "$1" = "stop" ]; then exit 0;
        else exit 5; fi; }

# Set up environment accordingly to the configuration settings
[ -n "$JENKINS_HOME" ] || { echo "JENKINS_HOME not configured";
        if [ "$1" = "stop" ]; then exit 0;
        else exit 6; fi; }
[ -d "$JENKINS_HOME" ] || { echo "JENKINS_HOME directory does not exist: $JENKINS_HOME";
        if [ "$1" = "stop" ]; then exit 0;
        else exit 1; fi; }

# Search usable Java as /usr/bin/java might not point to minimal version required by Jenkins.
# see http://www.nabble.com/guinea-pigs-wanted-----Hudson-RPM-for-RedHat-Linux-td25673707.html
candidates="
/etc/alternatives/java
/usr/lib/jvm/java-1.8.0/bin/java
/usr/lib/jvm/jre-1.8.0/bin/java
/usr/lib/jvm/java-1.7.0/bin/java
/usr/lib/jvm/jre-1.7.0/bin/java
/usr/bin/java
"
for candidate in $candidates
do
  [ -x "$JENKINS_JAVA_CMD" ] && break
  JENKINS_JAVA_CMD="$candidate"
done

JAVA_CMD="$JENKINS_JAVA_CMD $JENKINS_JAVA_OPTIONS -DJENKINS_HOME=$JENKINS_HOME -jar $JENKINS_WAR"
PARAMS="--logfile=/var/log/jenkins/jenkins.log --webroot=/var/cache/jenkins/war --daemon"
[ -n "$JENKINS_PORT" ] && PARAMS="$PARAMS --httpPort=$JENKINS_PORT"
[ -n "$JENKINS_LISTEN_ADDRESS" ] && PARAMS="$PARAMS --httpListenAddress=$JENKINS_LISTEN_ADDRESS"
[ -n "$JENKINS_HTTPS_PORT" ] && PARAMS="$PARAMS --httpsPort=$JENKINS_HTTPS_PORT"
[ -n "$JENKINS_HTTPS_KEYSTORE" ] && PARAMS="$PARAMS --httpsKeyStore=$JENKINS_HTTPS_KEYSTORE"
[ -n "$JENKINS_HTTPS_KEYSTORE_PASSWORD" ] && PARAMS="$PARAMS --httpsKeyStorePassword='$JENKINS_HTTPS_KEYSTORE_PASSWORD'"
[ -n "$JENKINS_HTTPS_LISTEN_ADDRESS" ] && PARAMS="$PARAMS --httpsListenAddress=$JENKINS_HTTPS_LISTEN_ADDRESS"
[ -n "$JENKINS_DEBUG_LEVEL" ] && PARAMS="$PARAMS --debug=$JENKINS_DEBUG_LEVEL"
[ -n "$JENKINS_HANDLER_STARTUP" ] && PARAMS="$PARAMS --handlerCountStartup=$JENKINS_HANDLER_STARTUP"
[ -n "$JENKINS_HANDLER_MAX" ] && PARAMS="$PARAMS --handlerCountMax=$JENKINS_HANDLER_MAX"
[ -n "$JENKINS_HANDLER_IDLE" ] && PARAMS="$PARAMS --handlerCountMaxIdle=$JENKINS_HANDLER_IDLE"
[ -n "$JENKINS_ARGS" ] && PARAMS="$PARAMS $JENKINS_ARGS"

if [ "$JENKINS_ENABLE_ACCESS_LOG" = "yes" ]; then
    PARAMS="$PARAMS --accessLoggerClassName=winstone.accesslog.SimpleAccessLogger --simpleAccessLogger.format=combined --simpleAccessLogger.file=/var/log/jenkins/access_log"
fi

exec $JAVA_CMD $PARAMS
```

jtnord:
I was working with upstart more than systemd.

As the deb is also used by Debain and ubuntu 14.04 LTS is still kicking around how will this work with installing there?  I have a feeling that the dependency may cause the startup to fail - but have not checked.

My personal preference here is to stop making the platform packages all things for everything and start doing specific ones for platforms (ubuntu 14.04, ubuntu 16.04, debian) and use upstart/systemd/sysVinit as appropriate and also put the files in the correct place to begin with (ie LSB compliant).  (I've said this multiple times that currently the installers are not really fit for purpose and effort needs to be put in to completely rewrite them from scratch.... :-( )

deb/build/debian/control
```diff
@@ -2,7 +2,7 @@ Source: @@ARTIFACTNAME@@
 Section: devel
 Priority: extra
 Maintainer: @@AUTHOR@@
-Build-Depends: debhelper (>= 9)
+Build-Depends: debhelper (>= 9), dh-systemd
```
jtnord: I have a feeling this may break Ubuntu 14.04?

deb/build/debian/jenkins.service
```diff
@@ -0,0 +1,24 @@
...
+[Service]
+Type=simple
+User=jenkins
+Group=jenkins
+SyslogIdentifier=jenkins
+UMask='0002'
+LimitNOFILE='8192'
+
+Environment=JENKINS_HOME=/var/lib/jenkins
+ExecStart=/usr/bin/java \
```
jtnord: a lot of other configuration that is in the sysv init is now missing (making things configurable via a file in `/etc/default/jenkins.default`)

deb/build/debian/jenkins.service
```diff
@@ -0,0 +1,24 @@
+[Unit]
+Description=Jenkins Continuous Integration Server
+After=network.target
+
+[Service]
+Type=simple
+User=jenkins
+Group=jenkins
+SyslogIdentifier=jenkins
+UMask='0002'
```
jtnord: this must to be 0007 as a minimum (normally would be 0027) otherwise your Jenkins secrets could be read by users that should have no access to them.

deb/build/debian/jenkins.service
```diff
@@ -0,0 +1,24 @@
...
+
+Environment=JENKINS_HOME=/var/lib/jenkins
+ExecStart=/usr/bin/java \
+    -jar /usr/share/jenkins/jenkins.war \
+    --webroot=/var/cache/jenkins/war \
+    --pluginroot=/var/cache/jenkins/plugins \
+    --httpPort='8080' \
+
+# We send SIGTERM(15), java exit code is 143 = 128 + 15
+SuccessExitStatus=143
```
jtnord: really - jenkins sucessfull exit should be zero - seems like something is wrong here.

Most likely this should not be using `SIG_TERM` but `SIG_INT`.

deb/build/debian/control
hashar: Given Jenkins 2.54 requires Java 8 which is not available on Ubuntu Trusty 14.04, I guess that can be discarded. The point stand for other .deb based distributions that might not have systemd.

deb/build/debian/jenkins.service
hashar: Jenkins makes `$JENKINS_HOME/secrets` user only regardless of umask.  Note that `deb/build/debian/jenkins.default ` does not set any umask at all though it suggests `027`.

deb/build/debian/jenkins.service
hashar: I definitely agree. The reason I have not reused the `deb/build/debian/jenkins.default` provided by the package is that it is meant to be interpreted by a shell when systemd only supports `KEY=Value`.

I guess an other file should be provided, I am not sure what dh_systemd recommends for that though.

deb/build/debian/jenkins.service
hashar: Jenkins does not handle Unix signals, that ends up killing the application and Java exit code is `128 + signal number`.   Killing it with a `SIGINT` would just cause java to exit with `130`.

The shell based script `deb/build/debian/jenkins.init` uses `daemon --stop` which emits `SIGTERM`.

deb/build/debian/control
jtnord: You can use the Oracle JDK 8 on that version without issue.
(hence partly why the init script has such a convoluted way of finding which java to use)

deb/build/debian/jenkins.service
jtnord: you can have secrets in places other than secrets... e.g. a `Secret` in a jobs `config.xml` where access to that can allow you to use it elsewhere and recover the secret for a job that you should have no access to.

deb/build/debian/jenkins.service
jtnord: sending SIGINT should cause the JVM to initiate a shutdown - and registered shutdown hooks are kicked off.

You can see this when you run the command locally and then hit `ctrl+c`  (and ctrl+c should be the same as SIG_INT if memory serves correctly)

deb/build/debian/control
paladox: I think there should be a branch created just for trusty.

Though trusty supports systemd, it's not installed by default and is buggy.

lifeofguenter:
+1 would love to see systemd being used - might need a `Restart=on-failure` somewhere?

Felixoid:
Is there any chance to have it merged? It's already late 2020 and every supported Ubuntu and Debian use systemd by default. Debian Buster even has an issue with recognition if the daemon is running or not since `/usr/bin/daemon` does not create a pidfile anymore.

timja:
> Is there any chance to have it merged? It's already late 2020 and every supported Ubuntu and Debian use systemd by default. Debian Buster even has an issue with recognition if the daemon is running or not since `/usr/bin/daemon` does not create a pidfile anymore.

the feature yes, this pull request has a number of unresolved issues as can be seen in James' review.
Ideally it would be done done for all packaging targets as well not just debian, but could be a separate pull request

bigon:
Just chiming in, you should probably either call the systemd sequence (add `--with systemd` when calling `dh`) in `debian/rules` file or increase the debhelper compatibility (in `debian/compat`) to at least version 10, note that the later might change other things in the package.

basil:
This PR was an excellent start.

#266 takes it to the finish line.

Please help me test it.

basil:
Closing in favor of #266. Many thanks for your contributions in this area, @paladox and @hashar! They were very helpful.

PR #94, opened 2017-04-06
Create needed paths for MSI and OSX packages deployment

mafraba:
Create the needed paths in MSI and OSX before deploying the same way it's done for the rest of the packages.

@reviewbybees esp @armfergom


oleg-nenashev:
Is there a JIRA issue for it?

PR #95, opened 2017-04-13
[JENKINS-43495] - Updated Java dependency to Java 8

oleg-nenashev:
Tested the behavior is correct on:

* openjdk-8-jdk - ok
* openjdk-8-jre - ok
* Oracle JDK 8 /latest  ok
* openjdk-7-jdk - fails (as designed)
* openjdk-7-jre - fails (as designed)

https://issues.jenkins-ci.org/browse/JENKINS-43495

@reviewbybees @batmat

### DIFF
--- a/deb/build/debian/control
+++ b/deb/build/debian/control
@@ -8,7 +8,7 @@ Homepage: @@HOMEPAGE@@
 
 Package: @@ARTIFACTNAME@@
 Architecture: all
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Yes, I raised the issue to Windows. There is also a problem with OpenSUSE
tests at least

oleg-nenashev:
@reviewbybees done

PR #96, opened 2017-04-13
Explain why the RPM dependencies are so wacky

svanoort:
If we can't fix it, and the hack (pushing a nasty virtual RPM to our own repo) is essentially worse... at least we can explain why it is the way it is. 

And the reason is Oracle.  

@reviewbybees

### DIFF
--- a/suse/build/SPECS/jenkins.spec
+++ b/suse/build/SPECS/jenkins.spec
@@ -17,19 +17,11 @@ URL:		@@HOMEPAGE@@
 Group:		Development/Tools/Building
 License:	@@LICENSE@@
 BuildRoot:	%{_tmppath}/build-%{name}-%{version}
-# see the comment below from java-1.6.0-openjdk.spec that explains this dependency
-# java-1.5.0-ibm from jpackage.org set Epoch to 1 for unknown reasons,
-# and this change was brought into RHEL-4.  java-1.5.0-ibm packages
-# also included the epoch in their virtual provides.  This created a
-# situation where in-the-wild java-1.5.0-ibm packages provided "java =
-# 1:1.5.0".  In RPM terms, "1.6.0 < 1:1.5.0" since 1.6.0 is
-# interpreted as 0:1.6.0.  So the "java >= 1.6.0" requirement would be
-# satisfied by the 1:1.5.0 packages.  Thus we need to set the epoch in
-# JDK package >= 1.6.0 to 1, and packages referring to JDK virtual
-# provides >= 1.6.0 must specify the epoch, "java >= 1:1.6.0".
-#
-# java-1_6_0-sun provides this at least
-Requires:	java >= 1.6, procps
+# Unfortunately the Oracle Java RPMs do not register as providing anything (including "java" or "jdk")
+# So either we make a hard requirement on the OpenJDK or none at all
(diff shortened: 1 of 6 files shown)


suse/publish/gen.rb
```diff
@@ -24,6 +24,10 @@
 </pre>
 
 <p>
+You will need to explicitly install a Java runtime environment, because Oracle's Java RPMs are incorrect and fail to register as providing a java dependency.
```
daniel-beck: Could this list the minimum JRE by Jenkins version, or should we just say "Java 8 or GTFO"?

FTR:

* 2.54 (2017-04) and newer: Java 8
* 1.612 (2015-05) and newer: Java 7
* 1.520 (2013-06) and newer: Java 6

So let's plan the Java 9 dependency for March 2019 :-)

oleg-nenashev: The proposal from @daniel-beck LGTM

suse/build/SPECS/jenkins.spec
```diff
@@ -17,19 +17,11 @@ URL:		@@HOMEPAGE@@
...
-# satisfied by the 1:1.5.0 packages.  Thus we need to set the epoch in
-# JDK package >= 1.6.0 to 1, and packages referring to JDK virtual
-# provides >= 1.6.0 must specify the epoch, "java >= 1:1.6.0".
-#
-# java-1_6_0-sun provides this at least
-Requires:	java >= 1.6, procps
+# Unfortunately the Oracle Java RPMs do not register as providing anything (including "java" or "jdk")
+# So either we make a hard requirement on the OpenJDK or none at all
+# Only workaround would be to use a java virtual package, see https://github.com/keystep/virtual-java-rpm
+# Requires: java >= 1:1.8.0
```
oleg-nenashev: Not correct since we still package 1.7 (e.g. 2.46)

svanoort: Addressed by pegging the legacy build to a specific commit until our LTS baseline is on Java 8.

suse/publish/gen.rb
svanoort: Ehhhhhh, then we have yet another thing that must be done manually and remembered when we add Java 9 support.  And historically since this is a purely-manual process we miss things.

svanoort: But okay I guess.

oleg-nenashev:
Then it just needs a fix, which has been proposed by @daniel-beck

jtnord:
why do we still require a specific JVM in the debain package?  causes issues on Ubuntu 14.04 which is still supported by the vendor.

daniel-beck:
@jtnord Notably, this one is for the RPMs that don't require Java due to the crazy.

But yes, as I wrote in http://lists.jenkins-ci.org/pipermail/jenkins-infra/2017-April/001108.html :

> Another indication we should just remove the Java dependency, like in the RPMs, and let users figure this out themselves. It's not our problem if there's no sane way to declare a Java dependency for Debian packages.

svanoort:
@oleg-nenashev @daniel-beck Re-review when you guys get a chance?

svanoort:
@jtnord Addressed the DEB point in the https://issues.jenkins-ci.org/browse/JENKINS-43629

daniel-beck:
LGTM

PR #97, opened 2017-04-26
[JENKINS-43849] Sort RPM and DEV packages by date instead of by name

armfergom:
See [JENKINS-43849](https://issues.jenkins-ci.org/browse/JENKINS-43849)

@reviewbybees

### DIFF
--- a/bin/list-packages.rb
+++ b/bin/list-packages.rb
@@ -17,9 +17,9 @@ def list_packages(path,glob,rel)
         debs = [];
         sftp.dir.glob(path,glob) { |f| debs << f }
         
-        p=/^[^_-]+/
-        
-        debs.sort {|x,y| y.name.gsub(p,"")<=>x.name.gsub(p,"")}.each do |f|
+        versionRE=/^([^\d]+[-_])(.*)([-_].+)$/
+
+        debs.sort {|x,y| Gem::Version.new(y.name.match(versionRE).captures[1])<=>Gem::Version.new(x.name.match(versionRE).captures[1])}.each do |f|
             puts "<tr><td><a href='#{rel}/#{f.name}'>#{f.name}</a></td>"
             
             t=Time.at(f.attributes.mtime).strftime('%Y/%m/%d')


bin/list-packages.rb
```diff
@@ -17,9 +17,7 @@ def list_packages(path,glob,rel)
         debs = [];
         sftp.dir.glob(path,glob) { |f| debs << f }
         
-        p=/^[^_-]+/
-        
-        debs.sort {|x,y| y.name.gsub(p,"")<=>x.name.gsub(p,"")}.each do |f|
+        debs.sort {|x,y| Time.at(y.attributes.mtime)<=>Time.at(x.attributes.mtime)}.each do |f|
```
recena: Order by date can work with the most of the cases but an order by name, taking into account the version naming, would be better.

armfergom: Comparing the versions would be simple using:

```
 Gem::Version.new(a) <=> Gem::Version.new(b
```

which is available in the machine this scripts are executed, so all good.

However, if you ask me, this implies extracting the version number from the `name` using a regex and that may be fragile if the package naming strategy changes for some reason. I still think that ordering by date is the correct, most simple and logic approach.

I will implement the suggested approach anyway so that it can be discussed.

armfergom:
Thanks for the reviews. I have changed the implementation and we are now ordering by version. Please, hold off on merging as I am now performing some tests to be sure this doesn't break anything. Will let you know when this is ready for merging.

recena:
@armfergom How did you try this? Could you show a screenshot?

armfergom:
My previous comment:

> Please, hold off on merging as I am now performing some tests to be sure this doesn't break anything.

For now, I have tested with a basic Ruby script that tests the regex and group capturing works. That's all:

```
versionRE=/^([^\d]+[-_])(.*)([-_].+)$/
string="jenkins-_2.32.2_all.deb"
p string.match(versionRE).captures[1]
```

Output:

```
2.32.2
```

Tested with different outputs. But anyway, see my previous comment. Have not run the packaging scripts yet to test, but will and will update the PR with screenshots or whatever proves this works..

armfergom:
I confirm that the new ordering works as expected. IMHO, this is ready to merge.

recena:
@armfergom Could you share a screenshot of the HTML generated?

recena:
@daniel-beck Any chance to get this PR merged?

armfergom:
@svanoort any chance we can get this merged?

svanoort:
Ask I told @recena when he asked, @armfergom, include those screenshots (publicly) so others can see that we verified it works correctly, and I will merge.

armfergom:
@svanoort a test I have made locally.

## Before this PR
current ordering 1

## After this PR
new ordering 1

svanoort:
Thanks, looks reasonable to me.  Merging as a maintainer since this has had double reviews already.

PR #101, opened 2017-07-07
[JENKINS-45287] Debian packaging should not be satisfied by Java 9

raul-arabaolaza:
[JENKINS-45287](https://issues.jenkins-ci.org/browse/JENKINS-45287)

As stated in the ticket I believe the best approach here is to mimic what the RPM package actually does, drop the java dependency on the package but add a Java check in the init script so it only starts jenkins if we are on Java 8 and prints an appropriate error message if not.

Also I have removed the hardcoded java route, now the installer uses `which java` to try to locate the java executable

I have performed the following tests: (all using Vagrant machines)

* Ubuntu 16.04 with OpenJDK 7 installed and configured as default via package manager -> Service does not start and shows the corresponding error message about using an incorrect java version

* Ubuntu 16.04 with OpenJDK 8 installed and configured as default via package manager -> Service starts
* Debian Stretch With OpenJDK 8 installed and configured as default via package manager -> Service starts

* Ubuntu 16.04 with OpenJDK 9 installed and configured as default via package manager -> Service does not start and shows the corresponding error message about using an incorrect java version
* Debian Stretch with OpenJDK 9 installed and configured as default via package manager -> Service does not start and shows the corresponding error message about using an incorrect java version

* Ubuntu 16.04 with Oracle Java 8 installed and configured as default via package manager -> Service starts
* Debian Stretch With Oracle Java 8 installed and configured as default via package manager -> Service starts

* Ubuntu 16.04 with Oracle Java 9 installed and configured as default via package manager -> Service does not start and shows the corresponding error message about using an incorrect java version
* Debian Stretch with Oracle Java 9 installed and configured as default via package manager -> Service does not start and shows the corresponding error message about using an incorrect java version

* Debian Stretch with Oracle Java 9 manually installed via tar.gz file and update-alternatives in a location not in path -> Service does not start with no java found error message
* Debian Stretch with Oracle Java 9 manually installed via tar.gz file and update-alternatives in a location not in path but with a symbolic link in PATH -> Service does not start and shows the corresponding error message about using an incorrect java version

* Debian Stretch with Oracle Java 8 manually installed via tar.gz file and update-alternatives in a location not in path -> Service does not start with no java found error message
* Debian Stretch with Oracle Java 8 manually installed via tar.gz file and update-alternatives in a location not in path but with a symbolic link in PATH -> Service does not start and shows the corresponding error message about using an incorrect java version

Java not found error message:
```
vagrant@stretch:~$ sudo service jenkins status
● jenkins.service - LSB: Start Jenkins at boot time
   Loaded: loaded (/etc/init.d/jenkins; generated; vendor preset: enabled)
   Active: failed (Result: exit-code) since Fri 2017-07-07 15:42:03 GMT; 5s ago
     Docs: man:systemd-sysv-generator(8)
  Process: 1118 ExecStop=/etc/init.d/jenkins stop (code=exited, status=0/SUCCESS)
  Process: 1200 ExecStart=/etc/init.d/jenkins start (code=exited, status=1/FAILURE)

Jul 07 15:42:03 stretch systemd[1]: Starting LSB: Start Jenkins at boot time...
Jul 07 15:42:03 stretch jenkins[1200]: ERROR: No Java executable found in current PATH: /bin:/usr/bin:/sbin:/usr/sbin
Jul 07 15:42:03 stretch jenkins[1200]: If you actually have java installed on the system make sure the executable is in the aforementioned path and that 'which java' returns the java executable path
Jul 07 15:42:03 stretch systemd[1]: jenkins.service: Control process exited, code=exited status=1
Jul 07 15:42:03 stretch systemd[1]: Failed to start LSB: Start Jenkins at boot time.
Jul 07 15:42:03 stretch systemd[1]: jenkins.service: Unit entered failed state.
Jul 07 15:42:03 stretch systemd[1]: jenkins.service: Failed with result 'exit-code'.
```
Incorrect Java error message:
```
vagrant@stretch:~$ sudo service jenkins status
● jenkins.service - LSB: Start Jenkins at boot time
   Loaded: loaded (/etc/init.d/jenkins; generated; vendor preset: enabled)
   Active: failed (Result: exit-code) since Fri 2017-07-07 15:50:57 GMT; 5s ago
     Docs: man:systemd-sysv-generator(8)
  Process: 1369 ExecStop=/etc/init.d/jenkins stop (code=exited, status=0/SUCCESS)
  Process: 1481 ExecStart=/etc/init.d/jenkins start (code=exited, status=1/FAILURE)

Jul 07 15:50:57 stretch jenkins[1481]: Found an incorrect Java version
Jul 07 15:50:57 stretch jenkins[1481]: Java version found:
Jul 07 15:50:57 stretch jenkins[1481]: java version "9"
Jul 07 15:50:57 stretch jenkins[1481]: Java(TM) SE Runtime Environment (build 9+176)
Jul 07 15:50:57 stretch jenkins[1481]: Java HotSpot(TM) 64-Bit Server VM (build 9+176, mixed mode)
Jul 07 15:50:57 stretch jenkins[1481]: Aborting
Jul 07 15:50:57 stretch systemd[1]: jenkins.service: Control process exited, code=exited status=1
Jul 07 15:50:57 stretch systemd[1]: Failed to start LSB: Start Jenkins at boot time.
Jul 07 15:50:57 stretch systemd[1]: jenkins.service: Unit entered failed state.
Jul 07 15:50:57 stretch systemd[1]: jenkins.service: Failed with result 'exit-code'.
```

@reviewbybees specially @daniel-beck (as the issue reporter) and @svanoort

### DIFF
--- a/deb/build/debian/jenkins.init
+++ b/deb/build/debian/jenkins.init
@@ -25,6 +25,8 @@ SCRIPTNAME=/etc/init.d/$NAME
 DAEMON=/usr/bin/daemon
 DAEMON_ARGS="--name=$NAME --inherit --env=JENKINS_HOME=$JENKINS_HOME --output=$JENKINS_LOG --pidfile=$PIDFILE"
 
+JAVA_ALLOWED_VERSION="18"
+
 if [ -n "$UMASK" ]; then
     DAEMON_ARGS="$DAEMON_ARGS --umask=$UMASK"
 fi
@@ -40,6 +42,26 @@ fi
 # Exit if not supposed to run standalone
 if [ "$RUN_STANDALONE" = "false" ]; then
     echo "Not configured to run standalone" >&2
+    exit 1
+fi
+
+# Make sure there exists a java executable, it may not be allways the case
+if [ -z "$JAVA" ]; then
+    echo "ERROR: No Java executable found in the system" >&2
+    echo "If you actually have java installed on the system make sure 'which java' returns the java executable path" >&2
+    exit 1
+fi
+
+# Work out the JAVA version we are working with:
+JAVA_VERSION=$($JAVA -version 2>&1 | sed -n ';s/.* version "\(.*\)\.\(.*\)\..*"/\1\2/p;')
+
+if [ "$JAVA_VERSION" = "$JAVA_ALLOWED_VERSION" ]; then
+    echo "Correct java version found" >&2
+else
+    echo "Found an incorrect Java version" >&2
+    echo "Java version found:" >&2
+    echo $($JAVA -version) >&2
+    echo "Aborting" >&2
     exit 1
 fi
 
--- a/deb/build/debian/jenkins.init
+++ b/deb/build/debian/jenkins.init
@@ -47,8 +47,8 @@ fi
 
 # Make sure there exists a java executable, it may not be allways the case
 if [ -z "$JAVA" ]; then
-    echo "ERROR: No Java executable found in the system" >&2
-    echo "If you actually have java installed on the system make sure 'which java' returns the java executable path" >&2
+    echo "ERROR: No Java executable found in current PATH: $PATH" >&2
+    echo "If you actually have java installed on the system make sure the executable is in the aforementioned path and that 'which java' returns the java executable path" >&2
     exit 1
 fi
 
--- a/deb/build/debian/jenkins.init
+++ b/deb/build/debian/jenkins.init
@@ -49,7 +49,7 @@ fi
 # Make sure there exists a java executable, it may not be allways the case
 if [ -z "$JAVA" ]; then
     echo "ERROR: No Java executable found in current PATH: $PATH" >&2
-    echo "If you actually have java installed on the system make sure the executable is in the aforementioned path and that 'which java' returns the java executable path" >&2
+    echo "If you actually have java installed on the system make sure the executable is in the aforementioned path and that 'type -p java' returns the java executable path" >&2
     exit 1
 fi
 
--- a/deb/build/debian/jenkins.init
+++ b/deb/build/debian/jenkins.init
@@ -24,6 +24,7 @@ SCRIPTNAME=/etc/init.d/$NAME
 #DAEMON=$JENKINS_SH
 DAEMON=/usr/bin/daemon
 DAEMON_ARGS="--name=$NAME --inherit --env=JENKINS_HOME=$JENKINS_HOME --output=$JENKINS_LOG --pidfile=$PIDFILE"
+JAVA=`type -p java`
 
 JAVA_ALLOWED_VERSION="18"
 
--- a/deb/publish/gen.rb
+++ b/deb/publish/gen.rb
@@ -31,6 +31,20 @@ Then add the following entry in your <tt>/etc/apt/sources.list</tt>:
 <pre style="padding:0.5em; margin:1em; background-color:black; color:white">
 deb #{url} binary/
 </pre>
+</p>
+
+<p>
+You will need to explicitly install a Java runtime environment, given the current state of Java 9 is the safest way to
+ensure your system ends properly configured. Adding an explicit dependency requirement on Java could force installation
+of undesired versions of the JVM.
+</p>
+
+<p>
+<ul>
+  <li>2.54 (2017-04) and newer: Java 8</li>
+  <li>1.612 (2015-05) and newer: Java 7</li>
+</ul>
+</p>
 
 <p>
 Update your local package index, then finally install #{productName}:
@@ -39,6 +53,7 @@ Update your local package index, then finally install #{productName}:
 sudo apt-get update
 sudo apt-get install #{artifactName}
 </pre>
+</p>
 
 <p>
 See <a href="http://wiki.jenkins-ci.org/display/JENKINS/Installing+Jenkins+on+Ubuntu">Wiki</a> for more information, including notes regarding upgrade from Hudson.
--- a/deb/publish/gen.rb
+++ b/deb/publish/gen.rb
@@ -34,7 +34,7 @@ deb #{url} binary/
 </p>
 
 <p>
-You will need to explicitly install a Java runtime environment, given the current state of Java 9 is the safest way to
+You will need to explicitly install a Java runtime environment, because Java 9 does not work with Jenkins, this is the safest way to
 ensure your system ends properly configured. Adding an explicit dependency requirement on Java could force installation
 of undesired versions of the JVM.
 </p>
--- a/deb/publish/gen.rb
+++ b/deb/publish/gen.rb
@@ -36,7 +36,8 @@ deb #{url} binary/
 <p>
 You will need to explicitly install a Java runtime environment, <b>because Jenkins does not work with Java 9</b>, this is the safest way to
 ensure your system ends properly configured. Adding an explicit dependency requirement on Java could force installation
-of undesired versions of the JVM.
+of undesired versions of the JVM. Check <a href="https://issues.jenkins-ci.org/browse/JENKINS-40689">JENKINS-40689</a>
+for more details about Jenkins and Java 9 compatibility.
 </p>
 
 <p>
--- a/deb/publish/gen.rb
+++ b/deb/publish/gen.rb
@@ -34,7 +34,7 @@ deb #{url} binary/
 </p>
 
 <p>
-You will need to explicitly install a Java runtime environment, because Java 9 does not work with Jenkins, this is the safest way to
+You will need to explicitly install a Java runtime environment, <b>because Jenkins does not work with Java 9</b>, this is the safest way to
 ensure your system ends properly configured. Adding an explicit dependency requirement on Java could force installation
 of undesired versions of the JVM.
 </p>
--- a/deb/build/debian/control
+++ b/deb/build/debian/control
@@ -8,7 +8,7 @@ Homepage: @@HOMEPAGE@@
 
 Package: @@ARTIFACTNAME@@
 Architecture: all
-Depends: ${misc:Depends}, daemon, adduser, procps, psmisc, net-tools, default-jre-headless (>= 2:1.8) | java8-runtime-headless
+Depends: ${misc:Depends}, daemon, adduser, procps, psmisc, net-tools
 Conflicts: hudson
 Replaces: hudson
 Description: @@DESCRIPTION_FILE@@
--- a/deb/build/debian/jenkins.default
+++ b/deb/build/debian/jenkins.default
@@ -4,7 +4,7 @@
 NAME=@@ARTIFACTNAME@@
 
 # location of java
-JAVA=/usr/bin/java
+JAVA=$(which java)
 
 # arguments to pass to java
 
--- a/deb/build/debian/jenkins.default
+++ b/deb/build/debian/jenkins.default
@@ -3,9 +3,6 @@
 # pulled in from the init script; makes things easier.
 NAME=@@ARTIFACTNAME@@
 
-# location of java
-JAVA=$(which java)
-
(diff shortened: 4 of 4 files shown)


deb/build/debian/jenkins.init
```diff
@@ -43,6 +45,26 @@ if [ "$RUN_STANDALONE" = "false" ]; then
...
+if [ -z "$JAVA" ]; then
+    echo "ERROR: No Java executable found in current PATH: $PATH" >&2
+    echo "If you actually have java installed on the system make sure the executable is in the aforementioned path and that 'which java' returns the java executable path" >&2
+    exit 1
+fi
+
+# Work out the JAVA version we are working with:
+JAVA_VERSION=$($JAVA -version 2>&1 | sed -n ';s/.* version "\(.*\)\.\(.*\)\..*"/\1\2/p;')
+
+if [ "$JAVA_VERSION" = "$JAVA_ALLOWED_VERSION" ]; then
```
oleg-nenashev: I am not sure it works correctly for all Java versios. Have you checked IBM Java for example?

samrocketman:
> now the installer uses `which java` to try to locate the java executable

I recommend `type -p java`.  It is a shell built in and more reliable than `which` in my experience.

daniel-beck:
> mimic what the RPM package actually does, drop the java dependency on the package

Note that the other package repo index files explain this and list the required Java versions, this should also be done for Debian then.

Compare the description on top of https://pkg.jenkins.io/redhat/ to https://pkg.jenkins.io/debian/

deb/build/debian/jenkins.init
raul-arabaolaza: @oleg-nenashev Yes, it works perfectly with IBM java 8:

{noformat}
ubuntu@ubuntu-xenial:~$ java -version
java version "1.8.0"
Java(TM) SE Runtime Environment (build pxa6480sr4fp7-20170627_02(SR4 FP7))
IBM J9 VM (build 2.8, JRE 1.8.0 Linux amd64-64 Compressed References 20170616_352529 (JIT enabled, AOT enabled)
J9VM - R28_20170616_0201_B352529
JIT  - tr.r14.java_20170616_352529
GC   - R28_20170616_0201_B352529_CMPRSS
J9CL - 20170616_352529)
JCL - 20170622_01 based on Oracle jdk8u131-b11
ubuntu@ubuntu-xenial:~$ sudo service jenkins start
ubuntu@ubuntu-xenial:~$ sudo service jenkins status
● jenkins.service - LSB: Start Jenkins at boot time
   Loaded: loaded (/etc/init.d/jenkins; bad; vendor preset: enabled)
   Active: active (exited) since Mon 2017-07-10 10:22:39 UTC; 5s ago
     Docs: man:systemd-sysv-generator(8)
  Process: 6285 ExecStop=/etc/init.d/jenkins stop (code=exited, status=0/SUCCESS)
  Process: 6386 ExecStart=/etc/init.d/jenkins start (code=exited, status=0/SUCCESS)

Jul 10 10:22:38 ubuntu-xenial systemd[1]: Starting LSB: Start Jenkins at boot time...
Jul 10 10:22:38 ubuntu-xenial jenkins[6386]: Correct java version found
Jul 10 10:22:38 ubuntu-xenial jenkins[6386]:  * Starting Jenkins Automation Server jenkins
Jul 10 10:22:38 ubuntu-xenial su[6427]: Successful su for jenkins by root
Jul 10 10:22:38 ubuntu-xenial su[6427]: + ??? root:jenkins
Jul 10 10:22:38 ubuntu-xenial su[6427]: pam_unix(su:session): session opened for user jenkins by (uid=0)
Jul 10 10:22:39 ubuntu-xenial jenkins[6386]:    ...done.
Jul 10 10:22:39 ubuntu-xenial systemd[1]: Started LSB: Start Jenkins at boot time.
{noformat}

Is IBM one of the supported JVM for jenkins? Not sure about this but, judging by the dependencies in the control file, IBM Java alone would not allow you to install the deb file. You would end having to install openjdk or the platform's default JVM which I suspect is not IBM on any case

raul-arabaolaza:
@samrocketman I am trying, but it breaks the script, If I try `JAVA=$(type -p java)` in default file I end with process existing before being able to install (type -p java) works on command line it seems is the evaluation as string what is making it fail. Some suggestion on how can I replace `JAVA=$(wich java)` with the type builtin?

deb/build/debian/jenkins.init
oleg-nenashev: Well, Jenkins project does not define which JDKs it supports. We do not test with IBM Java, but one may run with it if he wants. Some known issues: https://issues.jenkins-ci.org/issues/?jql=labels%20%3D%20java-ibm

raul-arabaolaza: Good to know, I have checked that the java detection changes in the init script work with IBM java 8. as long as the script is able to locate java in the classpath. This is the same situation than with custom installed versions of Oracle or OpenJDK runtimes, and an improvement over the previous situation where Java's path was simply hardcoded as `usr/bin/java`

raul-arabaolaza:
@daniel-beck Updated gen site index to state the need to install Java manually.

raul-arabaolaza:
@daniel-beck @samrocketman @oleg-nenashev I believe I have addressed your comments

deb/build/debian/jenkins.init
```diff
@@ -43,6 +46,26 @@ if [ "$RUN_STANDALONE" = "false" ]; then
     exit 1
 fi
 
+# Make sure there exists a java executable, it may not be allways the case
+if [ -z "$JAVA" ]; then
```
oleg-nenashev: IMHO `JAVA_HOME` must be checked as well

raul-arabaolaza: True, good point!

raul-arabaolaza:
@oleg-nenashev Can you point me to specific JVM implementations/versions outputting that version schemes you comment? I mean, you are right but unless  I can find a real JVM outputting them I am reluctant to complicate the regular expression just for something that is not proven real.

daniel-beck:
Notably, a VM that can actually run Jenkins. Meaning, I doubt J9 counts.

oleg-nenashev:
@daniel-beck IBM J9 is Java 8. Really

daniel-beck:
@oleg-nenashev 

> IBM J9 is Java 8. Really

I meant it has so many issues that I doubt anyone should be running this.

raul-arabaolaza:
@oleg-nenashev I have tested on IBM JVM and the current code works.

deb/build/debian/jenkins.init
samrocketman: Won't Jenkins just gracefully use `$USER/.jenkins` if `$JENKINS_HOME` is not set?

samrocketman:
> @samrocketman I am trying, but it breaks the script, If I try JAVA=$(type -p java) in default file I end with process exiting before being able to install (type -p java) works on command line it seems is the evaluation what is making it fail. Some suggestion on how can I replace JAVA=$(wich java) with the type builtin?

Can you elaborate?  `JAVA=$(type -p java)` should be evaluating just fine.  How are you encountering it?  Can you use `set -x` to debug where the issue occurs?

deb/build/debian/jenkins.init
raul-arabaolaza: @samrocketman I believe @oleg-nenashev is referring to use `JAVA_HOME` as location of the java executable if present

raul-arabaolaza:
@samrocketman `JAVA=$(type -p java)` was failing with `command not found` when included on the default file (which is sourced by the init script) but works perfectly on the init script itself. Is the same error than when I try to execute `sudo type -p java`. Not a real problem because I can search the Java path on the init script, not sure if I can consider a dynamic non overridable path a default, But I am curious why is happening.

raul-arabaolaza:
@oleg-nenashev The service command strips most env variables from the env of the init script is executed, including (if set `JAVA_HOME`), that means I can not access `JAVA_HOME` from the init script easily.

I could try to search and souce files like `/etc/profile` or `/etc/environment` but It feels like a hack and error prone. TBH that could complicate in such a way this PR for something that is not the main objective (JAVA_HOME was not used before) that is not something I would like to do.

Removing the hardcoded `/usr/bin/java` was relatively easy. But I do not want that to be the main objective of this PR, we can allways file a separate issue about honoring `JAVA_HOME` in all installers

oleg-nenashev:
> TBH that could complicate in such a way this PR for something that is not the main objective (JAVA_HOME was not used before) that is not something I would like to do.

Agreed

oleg-nenashev:
If this change is subject for backporting, I would rather advice not to backport it. Java 9 is not going to be released before the September LTS, hence we can keep it as is

samrocketman:
@raul-arabaolaza I was able to reproduce it in a different way.  Example,

```
$ echo $(type -p true)

$ type -t true
builtin
$ echo $(type -P true)
/bin/true
```

From the bash man page (emphasis mine),

>  type [-aftpP] name [name ...]
              With  no options, indicate how each name would be interpreted if
              used as a command name.  If the -t option is used, type prints a
              string  which  is  one  of alias, keyword, function, builtin, or
              file if  name  is  an  alias,  shell  reserved  word,  function,
              builtin,  or disk file, respectively.  If the name is not found,
              then nothing  is  printed,  and  an  exit  status  of  false  is
              returned.   If  the  -p  option is used, type either returns the
              name of the disk file that would be executed if name were speci‐
              fied as a command name, or nothing if `type -t name` would not
              return file.  **The -P option forces a PATH search for each  name,
              even if `type -t name` would not return file.**  If a command is
              hashed, -p and -P print the hashed value, which is not necessar‐
              ily  the  file  that appears first in PATH.  If the -a option is
              used, type prints all of the places that contain  an  executable
              named name.  This includes aliases and functions, if and only if
              the -p option is not also used.  The table of hashed commands is
              not  consulted  when  using  -a.  The -f option suppresses shell
              function lookup, as with the command builtin.  type returns true
              if all of the arguments are found, false if any are not found.

So the more reliable method I was thinking about was probably:

    JAVA=$(type -P java)

raul-arabaolaza:
@samrocketman In the end that was what I used but with backticks in the init script instead of the default file...

@daniel-beck As the issue creator it would be great a confirmation that you are ok with this changes or the approach.
@svanoort I believe you are the maintainer?

daniel-beck:
Seems reasonable given the terrible world we're living in 😭

deb/publish/gen.rb
```diff
@@ -31,6 +31,20 @@
 <pre style="padding:0.5em; margin:1em; background-color:black; color:white">
 deb #{url} binary/
 </pre>
+</p>
+
+<p>
+You will need to explicitly install a Java runtime environment, given the current state of Java 9 is the safest way to
```
svanoort: 🐛  I think you mean to say "because Java 9 does not work with Jenkins, this is the safest way" -- this sounds like Java 9 IS supported.

raul-arabaolaza:
@svanoort Documentation changed, good catch!

daniel-beck:
While we're at it, perhaps highlight in bold on these pages that Jenkins currently does not work with Java 9 (or at least hasn't been tested with it)? This would set expectations properly, i.e. "there be dragons".

raul-arabaolaza:
@daniel-beck Highligthed in bold the `because Java 9 does not work with Jenkins` part. Also I have changed it to `because Jenkins does not work with Java 9`

svanoort:
+1 for bolding that.

oleg-nenashev:
@raul-arabaolaza Makes sense to reference this ticket in the documentation then: https://issues.jenkins-ci.org/browse/JENKINS-40689

raul-arabaolaza:
@oleg-nenashev Done :)

PR #107, opened 2018-05-17
JENKINS-51253: add back java8 dependency to .deb packages

mostynb:
JENKINS-45287 removed java dependencies from .deb packages in an attempt
to not install the wrong JRE version (Jenkins currently requires Java 8),
and some distributions probably ship Java 9 now.  But this will break
existing Jenkins installs if the maintainer runs `apt-get autoremove`.

So let's add back a Java 8-specific dependency.

### DIFF
--- a/deb/build/debian/control
+++ b/deb/build/debian/control
@@ -8,7 +8,7 @@ Homepage: @@HOMEPAGE@@
 
 Package: @@ARTIFACTNAME@@
 Architecture: all
-Depends: ${misc:Depends}, daemon, adduser, procps, psmisc, net-tools
+Depends: ${misc:Depends}, daemon, adduser, procps, psmisc, net-tools, openjdk-8-jre-headless | oracle-java8-jre | jenkins-suitable-java8-jre
 Conflicts: hudson
 Replaces: hudson
 Description: @@DESCRIPTION_FILE@@
--- a/deb/publish/gen.rb
+++ b/deb/publish/gen.rb
@@ -34,10 +34,37 @@ deb #{url} binary/
 </p>
 
 <p>
-You will need to explicitly install a Java runtime environment, <b>because Jenkins does not work with Java 9</b>, this is the safest way to
-ensure your system ends properly configured. Adding an explicit dependency requirement on Java could force installation
-of undesired versions of the JVM. Check <a href="https://issues.jenkins-ci.org/browse/JENKINS-40689">JENKINS-40689</a>
-for more details about Jenkins and Java 9 compatibility.
+Note that Jenkins currently requires a Java 8 JRE. Check <a href="https://issues.jenkins-ci.org/browse/JENKINS-40689">JENKINS-40689</a>
+for more details about Jenkins and Java 9 incompatibility.
+</p>
+
+<p>
+If you wish to use Oracle's Java 8 JRE with this .deb package, we recommend that you do the following (adjust accordingly if you need a 32bit JRE):
+<ol>
+  <li>apt-get install java-package</li>
+  <li>Download the Oracle Java SE Runtime Environment 8 .tar.gz file for Linux from <a href="http://www.oracle.com/technetwork/java/javase/downloads/index.html">http://www.oracle.com/technetwork/java/javase/downloads/index.html</a></li>
+  <li>Build a .deb package from the downloaded JRE: make-jpkg jre-8u181-linux-x64.tar.gz</li>
+  <li>Install the Oracle .deb package: dpkg -i oracle-java8-jre_8u181_amd64.deb</li>
+  <li>Finally, add the jenkins repository as above and apt-get install jenkins</li>
+</ol>
+</p>
+
+<p>
+If you wish to use some JRE other than OpenJDK 8 or Oracle Java SE Runtime Environment 8, install it as you would normally and create a .deb package that satisfies Jenkins' requirements:
+<ol>
+  <li>apt-get install equivs</li>
+  <li>cat << EOF > jenkins-suitable-java8-jre.control
+Package: jenkins-suitable-java8-jre
+Description: A fake JRE to satisfy Jenkins' packaging
+ A package, which can be installed to satisfy the Jenkins' JRE dependency
+ when you have a locally installed suitable JRE.
+
+EOF
+</li>
+  <li>equivs-build jenkins-suitable-java8-jre.control</li>
+  <li>dpkg -i jenkins-suitable-java8-jre_1.0_all.deb</li>
+  <li>Finally, add the jenkins repository as above and apt-get install jenkins</li>
+</ol>
 </p>
 
 <p>


mostynb:
@svanoort: please take a look...

daniel-beck:
After @samrocketman's comment I remember the corresponding RedHat documentation on https://pkg.jenkins.io/redhat-stable/ which says:

>  You will need to explicitly install a Java runtime environment, because Oracle's Java RPMs are incorrect and fail to register as providing a java dependency. Thus, adding an explicit dependency requirement on Java would force installation of the OpenJDK JVM. 

In fact, we publish a similar notification on https://pkg.jenkins.io/debian-stable/ :

>  You will need to explicitly install a Java runtime environment, because Jenkins does not work with Java 9, this is the safest way to ensure your system ends properly configured. Adding an explicit dependency requirement on Java could force installation of undesired versions of the JVM. Check JENKINS-40689 for more details about Jenkins and Java 9 compatibility. 

I wonder whether the specified dependency would result in the installation of Java 9 on systems having the package available.

raul-arabaolaza:
I have two concerns in this PR:
 * I do not know if the concerns that move us to remove the Java package are still there as @daniel-beck has already stated.

* Forcing any user to install OpenJDK to use Jenkins is a no go, IMO. Yeah, `autoremove` may not work in some cases but I consider much worse a forceful installation of a concrete JVM implementation in your system. At least we should try to detect if there exists a JVM in the system before installing it, if that is possible now

IIUC for this fix to solve the `autoremove` problem users should be forced to update their jenkins version, I am right?

And last, but not least, if getting back the dependency is the path to go. Should also other packages be updated to include the dependency? Also docs should be updated too

mostynb:
Updated the PR, it now specifies two alternative dependencies: one for openjdk-8-jre-headless (which is sufficient for jenkins in my testing, not sure what went wrong in @samrocketman's case), and one for oracle-java8-installer which seems to be the preferred way to use Oracle's JVM in debian/ubuntu (reference: https://launchpad.net/~webupd8team/+archive/ubuntu/java).

@daniel-beck: this PR only affects .deb based linux distributions AFAIK, unless you do something unusual like convert .deb packages to .rpm's.

daniel-beck:
> this PR only affects .deb based linux distributions AFAIK

Just pointing out that our RPM package handles this issue the same way.

PR is incomplete without adapting https://github.com/jenkinsci/packaging/blob/master/deb/publish/gen.rb#L39

raul-arabaolaza:
+1 on Sam's comment

mostynb:
Updated the PR to remove the documentation about needing to separately install a JRE8 package.  But perhaps it should be updated to include some of the details which follow...

@svanoort: IIUC this won't cause any PPA's to be added, so webupd8team will not be accidentally installed unless the admin has already added it- in which case, it should be no surprise.

The preferred way to install Oracle JRE on debian/ubuntu based systems is to download the tarball from Oracle, and use make-jpkg (https://wiki.debian.org/JavaPackage) to turn the tarball into a .deb package and then install it with dpkg.  If you do this, then you get a package named something like oracle-java8-jre_8u171_amd64.deb and which provides the java8-runtime and java8-runtime-headless meta pacakges.  If you do this and subsequently apt-get install jenkins, then no new JRE should be installed.

Alternatively, you could install JRE manually, and use the "equivs" package to pretend that a regular JRE .deb package is installed, and hope that the jenkins .deb configuration is able to find and use it.  But I do not recommend this- the make-jpkg method is much simpler if you must have Oracle JRE, and OpenJDK is even simpler and suitable for most people.

@daniel-beck: I have seen commercial linux software packages provide their own JRE, which they install in some custom location as a step in their custom package installer (which is essentially a shell script, not a .deb or .rpm package).  IMO this is not ideal, and I'm not sure if Oracle even allows this without some direct agreement beforehand.

For RPM based distributions, Oracle provides a .rpm file, but I'm not familiar enough about RPM to recommend a procedure for using this with Jenkins.

mostynb:
Changed the proposed dependencies so they refer to openjdk-8-jre-headless and oracle-java8-jre instead of the java8-runtime-headless virtual packages.  I am not aware of any other JRE's that are likely to be used on debian/ubuntu so perhaps this is sufficient?

samrocketman:
> I'm not sure if Oracle even allows this without some direct agreement beforehand.

Worth mentioning that Oracle is not the only proprietary JVM vendor.  There are other vendors such as Azule Systems Zing.

I still agree we should leave the JVM install separate.  Additionally, I think we should remove the auto-detection for Java and either rely on Linux PATH expansion (I.e. just `java`) or fall back to the admin setting JAVA_HOME.

svanoort:
Jenkins is mostly-there for running with Java 10, so... hard dependency on Java 8 is again, a no-go.

mostynb:
When Jenkins is java10-ready, the condition should be updated, of course.  Something like this might make sense at that point:
java10-runtime-headless | openjdk-8-jre-headless | oracle-java8-jre

Another option that could work now would be to depend upon a virtual package that doesn't exist, but which advanced users/sysadmins could create with the equivs package. eg the jenkins .deb could depend on something like:
openjdk-8-jre-headless | oracle-java8-jre | jenkins-N-suitable-jre (where N describes the jenkins version)

That way:
* Users that don't care which jre is used just apt-get install jenkins and get openjdk8.
* Users that require oracle would simply install that first, then apt-get install jenkins and it would be used.
* Users that require an obscure jre would install it, then create and install a dummy package with equivs which provides jenkins-N-suitable-jre, then apt-get install jenkins.

Of these three cases, the first two are probably most common, and are relatively simple.

samrocketman:
@svanoort it would be great if Jenkins core as part of its startup sequence of `Jenkins.instance` detected the version of Java and gave a helpful message in the error log.  Currently, core just attempts to use JVM features and then stack traces when it errors.  Which leaves a user up to web searching the stack trace and learning from stackoverflow or a Jenkins issue that they're using an unsupported version of Java.

I'm a big fan of surfacing common errors in as human readable / helpful form as possible.

Since developing for future JVMs will always be desired there could be a system property to skip JVM version checking for development in that scenario.  What are your thoughts on this?

daniel-beck:
@samrocketman It already should per https://github.com/jenkinsci/extras-executable-war/blob/master/src/main/java/Main.java#L126...L155 -- What version of Jenkins are you using, and how are you running it? File a bug with complete steps to reproduce. (Notably, if you install in Tomcat or similar, you're expected to read docs, so that doesn't count.)

samrocketman:
@daniel-beck I haven't seen it recently so was probably going off of an old memory.  I retract what I said.

daniel-beck:
@mostynb I think your comment outlines a feasible way forward (with a disclaimer that I know next to nothing about Linux package management).

If it's not too much work, could you update this PR, including the doc for https://pkg.jenkins.io/debian/ / https://pkg.jenkins.io/debian-stable/ (at https://github.com/jenkinsci/packaging/blob/master/deb/publish/gen.rb) to explain how option 3 works?

CC @svanoort WDYT?

mostynb:
@daniel-beck: updated the commit, but I haven't cross-checked it yet.  Can you clarify that this is the suggestion that you think is feasible?

PR #109, opened 2018-06-12
[FIXES JENKINS-51888] Fixed PID mix up bug in init script

njesper:
Improved PID handling in RedHat/CentOS init script to fix bug:
https://issues.jenkins-ci.org/browse/JENKINS-51888

### DIFF
--- a/rpm/build/SOURCES/jenkins.init.in
+++ b/rpm/build/SOURCES/jenkins.init.in
@@ -114,13 +114,7 @@ case "$1" in
                 # [ "$sess" = "$MY_SESSION_ID" ] || continue
 	       	echo "$cmd" | grep $JENKINS_WAR > /dev/null
 		[ $? = 0 ] || continue
-                echo "$cmd" | grep -e "$JENKINS_PORT" > /dev/null
-                [ $? = 0 ] || continue
-                echo "$cmd" | grep -F "$JENKINS_LISTEN_ADDRESS" > /dev/null
-                [ $? = 0 ] || continue
-                echo "$cmd" | grep -e "$JENKINS_HTTPS_PORT" > /dev/null
-                [ $? = 0 ] || continue
-                echo "$cmd" | grep -F "$JENKINS_HTTPS_LISTEN_ADDRESS" > /dev/null
+                echo "$cmd" | grep -e "$JENKINS_HOME" > /dev/null
                 [ $? = 0 ] || continue
 		# found a PID
 		echo $pid > "$JENKINS_PID_FILE"
--- a/rpm/build/SOURCES/jenkins.init.in
+++ b/rpm/build/SOURCES/jenkins.init.in
@@ -114,6 +114,14 @@ case "$1" in
                 # [ "$sess" = "$MY_SESSION_ID" ] || continue
 	       	echo "$cmd" | grep $JENKINS_WAR > /dev/null
 		[ $? = 0 ] || continue
+                echo "$cmd" | grep -e "$JENKINS_PORT" > /dev/null
+                [ $? = 0 ] || continue
+                echo "$cmd" | grep -F "$JENKINS_LISTEN_ADDRESS" > /dev/null
+                [ $? = 0 ] || continue
+                echo "$cmd" | grep -e "$JENKINS_HTTPS_PORT" > /dev/null
+                [ $? = 0 ] || continue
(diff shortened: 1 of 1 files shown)


daniel-beck:
This appears to introduce the problem that I cannot change the defaults like listen port, and have Jenkins restart afterwards, as the process would not be found, right?

njesper:
The problem right now is that when two masters are running from the same war-file under the same user, the script picks the last pid from the `/bin/ps hww -u "$JENKINS_USER" -o sess,ppid,pid,cmd` command. (Perhaps not an optimal way to do it, but it's a way to get updates to all masters from yum)

I haven't set up the package building environment and run a complete package and deploy test, but the added/modified lines uses variables from the `/etc/sysconfig/jenkins` in the same style as other parts of the script. E.g.:
```
  echo "$cmd" | grep $JENKINS_WAR > /dev/null
  [ $? = 0 ] || continue
```
at line 115-116 (although `$JENKINS_WAR` is set in the script, while e.g. `$JENKINS_HTTPS_PORT` at line 121 comes from the conf file). I'm running multiple masters using copies of the script and settings file, and it pics up the different ports without problems. Since I'm not using the @@-tokens, such as `@@PORT@@`, for the added filtering/identification I'm guessing these modifications will just pass through the package/deploy-process untouched and pick up config changes at every (re)start of Jenkins.

The fix should also cover the case if someone wants to set up different masters using the same port, but on different `LISTEN_ADDRESS`:es/IPs. I haven't tested this, however.

Are there other supported ways of changing the default settings? (I'm guessing adding the SSL-parameters in the `JENKINS_ARGS=` setting instead of the explicit parameters could be one way, but then e.g. `$JENKINS_HTTPS_PORT` would be empty and grep would exit with 0 and the script would behave as before.)

njesper:
Ah, now I get your first comment a little better, and I was just about to reply 'Yes, you are correct', but on second thoughts:

Line 117-124 just adds a fix to better identify the PID of what the `daemon` command at line 103 have started, before writing it to `$JENKINS_PID_FILE` (there are still some small holes if you run really esoteric configurations).

When stopping/killing or checking status, I've only added a parameter to actually use the pid-file (which the original script created, with potentially wrong content, and then never used). Thus stop/restart should be unaffected by changes to `/etc/sysconfig/jenkins` (unless you're also doing something with the pid-file, but users doing that are probably aware that it might result in unexpected behavior).

What might cause problems is running `service jenkins start` twice, but if I've interpreted the `daemon` function correctly, it should only do `exit 0` with the parameters the Jenkins init script is using, so this should also be safe.

njesper:
Ok, so I dug a bit deeper into the inner workings of /etc/init.d/functions and have to admit that the pid-file is used a bit more than I initially thought, albeit through implicit default behaviors. I don't see any problem with being explicit here (adding the `-p` parameter in `stop` and `status`).

I definitely agree that the sequence:
> 1. Start Jenkins
> 2. Edit `/etc/sysconfig/jenkins`
> 3. Stop Jenkins -- which won't work if the port was edited in step 2.

should continue to work. And, as far as I have now tested it, it does.

However, as I was getting close to at the end of the previous comment, running `service jenkins start` twice in this way:
1. Start Jenkins
2. Edit `/etc/sysconfig/jenkins`
3. Run `service jenkins start` 

will result in the pid-file being cleared in some cases, which would be a regression of functionality as well, although less common.

I'll be back with a more polished update shortly :-)
Thanks for your time so far, @daniel-beck

njesper:
Judging from the comments in the code in the start case of the init script, it seems to be known that the script can fail at finding the process/PID that it has just started. This update tries to improve that a little bit, while also improving the support for running multiple Jenkins masters in the same machine (JENKINS-51888).

Except for the trivial start/stop/status/... use cases two more intricate have been identified:
1. Start Jenkins
2. Edit config
3. Stop Jenkins 

and
1. Start Jenkins
2. Edit config
3. Run Jenkins start command again

Both cases will be handled by input validation/sanity checks during script startup (line 57-62) and the script will behave at least as good as before. ("Start|Edit|Start" will exit with an error, while "Start|Edit|Stop" actually will exit cleanly without doing anything.  - Perhaps the latter case should be looked into. An alternative behavior could be to check for the PID file and kill a matching process, if it exists.)

@daniel-beck, is this an acceptable update?

PR #110, opened 2018-07-09
Adding a lockfile so Jenkins shuts down correctly on RedHat-based distros

marcelocarlos:
RedHat-based (Centos, Amazon Linux, etc) distros expect a lockfile so it stops the services correctly when shutting down or rebooting.

The current init file is based on the Suse RPM version, which doesn't require that.

This updates the init script to create and remove the lockfile to ensure graceful shutdown/reboot of jenkins servers.

It also adds a chkconfig comment to the init file, so we can better define its runlevels.

Tested in Amazon Linux 1 (2018.03) and working as expected.

### DIFF
--- a/rpm/build/SOURCES/jenkins.init.in
+++ b/rpm/build/SOURCES/jenkins.init.in
@@ -1,6 +1,7 @@
 #!/bin/sh
 #
-#     SUSE system statup script for Jenkins
+#     RedHat system statup script for Jenkins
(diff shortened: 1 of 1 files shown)


marcelocarlos:
What's the next step before we can merge it? Anything pending?

oleg-nenashev:
I'd guess it needs an action from the repo maintainer.
CC @svanoort who is the maintainer AFAIK

PR #111, opened 2018-12-12
[FIXED JENKINS-55149] deb access set in JENKINS_ARGS

keithmarsh:
The options to enable access logs were set in DAEMON_ARGS, breaking daemon startup.  Moved access log options to append to JENKINS_ARGS.


oleg-nenashev:
Thanks for the PR @keithmarsh !

PR #113, opened 2019-03-06
[JENKINS-52284,JENKINS-55988,JENKINS-55989,JENKINS-55990,JENKINS-55991] - Update Unix and MacOS installers

oleg-nenashev:
It should fix all non-Windows installers, there is no critical patches. Just docs & Co

CC @jenkinsci/java11-support

### DIFF
--- a/deb/publish/gen.rb
+++ b/deb/publish/gen.rb
@@ -34,10 +34,10 @@ deb #{url} binary/
 </p>
 
 <p>
-You will need to explicitly install a Java runtime environment, <b>because Jenkins does not support all Java versions</b>, this is the safest way to
+You will need to explicitly install a Java Runtime Environment, <b>because Jenkins does not support all Java versions</b>, this is the safest way to
 ensure your system ends properly configured. Adding an explicit dependency requirement on Java could force installation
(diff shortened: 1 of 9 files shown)


deb/publish/gen.rb
```diff
@@ -34,14 +34,15 @@
 </p>
 
 <p>
-You will need to explicitly install a Java runtime environment, <b>because Jenkins does not work with Java 9</b>, this is the safest way to
+You will need to explicitly install a Java Runtime Environment, <b>because Jenkins does not support all Java versions</b>, this is the safest way to
```
oleg-nenashev: @MRamonLeon maybe even that. Oracle is merging JDK and JRE in its distributions, so maybe we do not want to keep JRE as a term anymore

```suggestion
You will need to explicitly install Java, <b>because Jenkins does not support all Java versions</b>, this is the safest way to
```

MRamonLeon: It looks definitively better

oleg-nenashev:
There is no CI, but it works on my machine (c). I believe @v1v has also tested the chage

PR #114, opened 2019-04-04
Fixed 'daemon: unrecognized option'

oomk:
In case of JENKINS_ENABLE_ACCESS_LOG="yes" options should be added into JENKINS_ARGS instead of DAEMON_ARGS. Otherwise 'daemon: unrecognized option' error occurs


oomk:
Closed, because it was done in https://github.com/jenkinsci/packaging/pull/111

PR #115, opened 2019-04-18
RPM: Add settings for HTTP/2 in sysconfig

tn-osimis:


tn-osimis:
Warning: Untested, just a suggestion.

tn-osimis:
@olamy Indeed. What do you think about [this](https://github.com/tn-osimis/packaging/pull/1/commits)? I believe something like it would suffice for Java 9+.

olamy:
@tn-osimis as I said in doc 
``` With Java9+ you do not need anymore to include the ALPN jar in the bootclasspath but only the jar jetty-alpn-java-server in the extraLibFolder  ``` :)

tn-osimis:
@olamy Oh not at all, I was indeed following this documentation. Just wanted to make sure everyone was fine with this being a Java 9+ thing only since I didn't include a setting for the `-Xbootclasspath/p` option (and fine with the general approach indeed; I guess this could be smarter about detecting versions and dependencies automatically but I decided against it).

So, I pulled the change in this PR accordingly. I'll try to build and test ASAP but if you preempt me I reckon this could be ready for merge.

PR #121, opened 2019-08-27
Add a Release Drafter configuration

oleg-nenashev:
This change reimplements Docker release notes from: https://github.com/jenkinsci/docker/releases .
I believe it might be helpful to have some kind of changelogs in this repo as well.

CC @jenkinsci/sig-platform @slide


oleg-nenashev:
I added some releases to https://github.com/jenkinsci/packaging/releases , but it is even more complicated than Docker packaging

PR #128, opened 2019-12-09
Make sure jenkins is really stopped before start in debian init script

tmad:
This should prevent running multiple Jenkins processes at once.

The problem was that killall doesn't check if the process was really stopped.
In case Jenkins process is stuck, daemon can get killed leaving Jenkins java process behind.
Upon startup the stuck process could be ignored, since it doesn't have to have a daemon process running and it doesn't need to listen on @@PORT@@ any more thus leading to two Jenkins java processes running

Still, following code might lead to an unexpected situation with systemd. Systemctl runs init script invoked with `restart` as two invoked with `stop` followed by `start`. When stop times out, it would exit with failure, but systemctl would still run the init script with `start` regardless of that failure.
Start would exit with success in case daemon process is still be running (hence restart would be successful), but:
* Jenkins could be just finishing the shutdown, or
* no Jenkins processes were really stopped and Jenkins is still running.
I think this would still be better to end up with that, instead of ending up with two of them, which might lead to an unexpected behaviour.

### DIFF
--- a/deb/build/debian/jenkins.init
+++ b/deb/build/debian/jenkins.init
@@ -135,6 +135,9 @@ do_start()
     #   2 if daemon could not be started
     $DAEMON $DAEMON_ARGS --running && return 1
 
+    # Verify if there is a jenkins process already running without a daemon
+    get_running || return 2
+
     # Verify that the jenkins port is not already in use, winstone does not exit
     # even for BindException
     check_tcp_port "http" "$HTTP_PORT" "@@PORT@@" "$HTTP_HOST" "0.0.0.0" || return 2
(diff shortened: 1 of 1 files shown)


olblak:
we could do both :) , unfortunately, I am out of power here

oleg-nenashev:
OK, leave it to me

tmad:
@oleg-nenashev I've change the destination branch to master

oleg-nenashev:
Thanks!

tmad:
fwiw I don't have write access to the repo, so I can't merge it on my own

oleg-nenashev:
Sorry, dropped the ball on it

PR #130, opened 2020-02-27
Add new releaseline callled experimental, more few fixes

olblak:
* Create a new releaseline called 'experimental' to not impact the current release process
* Only define BUILDENV, CREDENTIALS and BRAND variable if they are not set
* Refactor msi publishing script to read variable definition from makefile variable files


env/test.mk
```diff
@@ -13,8 +13,9 @@ export SCP_OPTS=
 # where to put binary files
 export TESTDIR=$(realpath .)/pkg.jenkins.io
 export WARDIR=/packages/binary/war${RELEASELINE}
+# Concat MSDIR and RELEASELINE in the msi publishing
 export MSIDIR=/packages/binary/windows${RELEASELINE}
-export OSXDIR=export OSXDIR=${TESTDIR}/osx${RELEASELINE}
+export OSXDIR=/packages/osx${RELEASELINE}
```
timja: is OSX still needed here? I thought we didn't ship it anymore

env/release.mk
```diff
@@ -12,7 +12,7 @@ export SCP_OPTS=-P 22
 
 # where to put binary files
 export WARDIR=/packages/binary/war${RELEASELINE}
-export MSIDIR=C:\packages\binary\windows${RELEASELINE}
+export MSIDIR=/packages/binary/windows${RELEASELINE}
```
timja: no longer on c:/ ?

env/test.mk
olblak: We don't ship it anymore and we could problaby remove those variables

env/release.mk
olblak: It appears that powershell is doing the translatation, at least it's working :p.

olblak:
I propose to do the OSX code cleanup in a different pull request as It would require some additionnal testing to be sure that part of it is not used somewhere else

PR #143, opened 2020-04-06
Remove array handling in SSH/SCP OPS

timja:


olblak:
I propose this solution instead #144

PR #144, opened 2020-04-06
Fix bash arrays when items need to be retrieved as separated word

olblak:
${array[*]}, returns every items as a single string where ${array[@]} return every items as separated components.


olblak:
https://github.com/koalaman/shellcheck/wiki/Sc2086

olblak:
I am going to move forward and this test this

PR #152, opened 2020-04-07
Link from Debian changelog to project changelog

MarkEWaite:
The current Debian changelog only provides the version number and user name of the account that ran the publishing script.  Let's link to the online changelog for those users that read the Debian changelog.

I've reviewed the [Debian changelog requirements](https://www.debian.org/doc/manuals/maint-guide/dreq.en.html#changelog) and the [Debian policy](https://www.debian.org/doc/debian-policy/ch-source.html#s-dpkgchangelog) and believe this change complies with the requirements and the policy.

@daniel-beck and @oleg-nenashev may also want to comment if this is a reasonable addition.

## Before:

```
jenkins (2.222.1) unstable; urgency=low

  * Packaged 2.222.1

 -- Kohsuke Kawaguchi <kk@kohsuke.org>  Wed, 25 Mar 2020 07:12:05 -0700

```
## After:

```
jenkins (2.222.1) unstable; urgency=low

  * Packaged 2.222.1 https://jenkins.io/changelog-stable/#v2.222.1

 -- Kohsuke Kawaguchi <kk@kohsuke.org>  Wed, 25 Mar 2020 07:12:05 -0700

```

### DIFF
--- a/deb/build/build.sh
+++ b/deb/build/build.sh
@@ -22,17 +22,10 @@ sed -i.bak -e 's/^\s*$/./' -e 's/^/ /' $DESCRIPTION_FILE
 # Rewrite the file
 mv "$DESCRIPTION_FILE.bak" "$DESCRIPTION_FILE"
 
-# Define changelog link for weekly releases
-CHANGELOG_LINK=https://jenkins.io/changelog/#v${VERSION}
-if [[ $VERSION =~ [0-9]+[.][0-9]+[.][0-9]+ ]]; then
-  # Link the LTS version to the LTS changelog
-  CHANGELOG_LINK=https://jenkins.io/changelog-stable/#v${VERSION}
-fi
-
 cat > $D/debian/changelog << EOF
 ${ARTIFACTNAME} ($VERSION${DEB_REVISION}) unstable; urgency=low
 
-  * Packaged ${VERSION} ${CHANGELOG_LINK}
+  * Packaged ${VERSION} https://jenkins.io/changelog${RELEASELINE}/#v${VERSION}
 
  -- ${AUTHOR}  $(date -R)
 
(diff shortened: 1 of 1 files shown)


deb/build/build.sh
```diff
@@ -22,10 +22,17 @@ sed -i.bak -e 's/^\s*$/./' -e 's/^/ /' $DESCRIPTION_FILE
 # Rewrite the file
 mv "$DESCRIPTION_FILE.bak" "$DESCRIPTION_FILE"
 
+# Define changelog link for weekly releases
+CHANGELOG_LINK=https://jenkins.io/changelog/#v${VERSION}
```
olblak, line 26:
```suggestion
CHANGELOG_LINK=https://jenkins.io/changelog${RELEASELINE}/#v${VERSION}
```
Wouldn't it be better to use RELEASELINE instead? My main concern is, as far as I know jenkins.io/changelog-* generation is not related to release process, so let image that we have an experimental release, the description will be broken

MarkEWaite: As far as I can tell, `RELEASELINE` has 5 allowed values (in order from most frequently used to least frequently used):

* (empty) - weekly
* stable
* stable-rc
* experimental
* rc

The changelogs are maintained on jenkins.io for two of those lines:

* (empty) - weekly
* stable

I'm fine with using RELEASELINE since it will remove logic and pattern matching from the script.

We have ongoing efforts to automate the changelog generation process.  Until the process is completely automated, the changelog link would be unsatisfied until the changelog is published.  

In the last several weeks, changelogs have been published within a few hours of the release or at the time of release.  In the last 6-12 months, the longest delay between Jenkins release and changelog publication has consistently happened within 1-2 days of release.

I think that a potentially broken link to a changelog is still significantly better than the current non-existing changelog information in the Debian/Ubuntu packaging.

MarkEWaite: See 8bb1331b4b15e2209c7cfc531e49b49c2374afc8 for the revision

PR #157, opened 2020-04-08
Add two protection to not overwrite existing packages

olblak:
* Rsync skip upload if remote file already exist
* Skip publishing script if remote file already exist


deb/publish/publish.sh
```diff
@@ -125,6 +155,7 @@ function signSite(){
 }
 
 show
+isPackagePublished
```
timja, line 158: What about skipIfAlreadyPublished?

PR #159, opened 2020-04-15
IndexGenerator python script now generate html for Apache fancy directory

olblak:
In the first attempt, the python script only generated a list of packages based on local packages which work for the testing environment but in the case where we already have plenty of packages on pkg.jenkins.io, in this case we would upload two files HEADER.html and FOOTER.html to the remote repository directory

If the apache configuration contains at least the following configuration, it inject some html code before (HEADER.html) and after (FOOTER.html) the directory listing

[Documentation](https://httpd.apache.org/docs/current/mod/mod_autoindex.html#adddescription) 

```
DocumentRoot "/usr/local/apache2/htdocs"
<Directory "/usr/local/apache2/htdocs">
    Options Indexes FollowSymLinks
    AllowOverride FileInfo
    Require all granted

    IndexOptions FancyIndexing VersionSort TrackModified FoldersFirst SuppressHTMLPreamble NameWidth=* SuppressRules SuppressIcon SuppressDescription

    HeaderName HEADER.html
    ReadmeName FOOTER.html

    IndexIgnore HEADER* FOOTER* index.html

</Directory>
```

Remark: Css work is just a proof of concept, but I leave it to another contribution

Debian screenshot

image


timja:
What's the easiest way to test this locally?

olblak:
> What's the easiest way to test this locally?

I added the docker-compose file that I used to test this so first you need to generate debian,redhat,etc packages. or you could mock them just `mkdir -p target/debian/ && touch target/debian/jenkins.deb` then you can generate the html files `./bin/indexGenerator.py -d debian -o target/debian` and finally curl http://localhost/debian

timja:
> > What's the easiest way to test this locally?
> 
> I added the docker-compose file that I used to test this so first you need to generate debian,redhat,etc packages. or you could mock them just `mkdir -p target/debian/ && touch target/debian/jenkins.deb` then you can generate the html files `./bin/indexGenerator.py -d debian -o target/debian` and finally curl http://localhost/debian

Thanks I used that for testing

olblak:
> I pushed a commit to fix the rendering I wasn't happy with, I've checked all the distributions,
> 
> only question is do we want the signing key block to be there on the windows and MSI pages? there's no information on how or if users should verify the signature, I assume windows handles it out of the box so its probably useless there, and not sure about the war one?

Just lack of time but I think it would be usefule to have instruction about verifying certificate, gpg k signature and sha256 if one is available like MSI

slide:
> > I pushed a commit to fix the rendering I wasn't happy with, I've checked all the distributions,
> > only question is do we want the signing key block to be there on the windows and MSI pages? there's no information on how or if users should verify the signature, I assume windows handles it out of the box so its probably useless there, and not sure about the war one?
> 
> Just lack of time but I think it would be usefule to have instruction about verifying certificate, gpg k signature and sha256 if one is available like MSI

Windows will automatically verify the signature of the MSI when installing.

PR #160, opened 2020-04-16
Remove unused parameter

timja:


deb/publish/publish.sh
```diff
@@ -30,7 +30,6 @@ function generateSite(){
   
   "$BASE/bin/indexGenerator.py" \
     --distribution debian \
-    --binaryDir "$DEBDIR" \
     --targetDir "$DEB_WEBDIR"
```
olblak, line 33: targetdIr should be be DEBDIR instead of DEB_WEBDIR, because it suppose to be working in the directory that contains the files, in facts we are really missing an index.html for website

olblak: I am going to merge this pull request is it's blocker and open a different one with my comment

PR #162, opened 2020-04-16
Various Jenkins publishing scripts fixes

olblak:
- [x] MSI needs to synchronize from a working directory to the remote location
- [x] MSI needs to publish generated html in the binary directory
- [x] WAR needs to synchronize from a working directory to the remote location
- [x] WAR needs to publish generated html in the binary directory
- [x] Debian repository metadata must be updated even if file already exist
- [x] Debian needs to publish generated html in the binary directory
- [x] RedHat repository metadata must be updated even if file already exist
- [x] RedHat needs to publish generated html in the binary directory
- [x] Suse needs to publish generated html in the binary directory

### DIFF
--- a/deb/publish/publish.sh
+++ b/deb/publish/publish.sh
@@ -164,7 +164,9 @@ function signSite(){
 }
 
 show
-skipIfAlreadyPublished
+## Disabling this function allow us to recreate and sign the repository.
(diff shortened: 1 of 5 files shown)


olblak:
I am currently running some test on a local environment

PR #169, opened 2020-04-17
Fix debian instructions

timja:


timja:
it's still broken but at least this will allow apt-get update to pass

PR #170, opened 2020-04-19
Remove staging from destination dir

timja:
I think this is the reason for pkg.jenkins.io website not being updated?


olblak:
Nope it's not that, you need to run the sync. Sh scripy

olblak:
The website is updated from this [script ](https://github.com/jenkins-infra/jenkins-infra/blob/staging/dist/profile/files/mirrorbrain/sync.sh#L68-L65), keep in mind that puppet doesn't run on this specific machine, so any change to that script currently need to be done directly on the machine

PR #171, opened 2020-04-20
Add site to pkg.jenkins.io staging

timja:


olblak:
This will copy HEADER.html and FOOTER.html to pkg.jenkins.io in order to be used with fancy index directory listing.
Because there is already an index.html, directory listing won't be enabled.

Instead I would suggest to merge to two files into one index.html

timja:
> This will copy HEADER.html and FOOTER.html to pkg.jenkins.io in order to be used with fancy index directory listing.
> Because there is already an index.html, directory listing won't be enabled.
> 
> Instead I would suggest to merge to two files into one index.html

why don't we remove the index.html and let fancy indexing take over?

PR #178, opened 2020-04-22
Generate index.html for pkgsite

olblak:
Generate index.html for pkg.jenkins.io based on header.html and footer.html


olblak:
Context was missing so index.html were not complete which is fixed now

PR #179, opened 2020-04-22
use IP intead of dns record for PKGSERVER 

olblak:
We are correctly experimenting Fastly for pkg.jenkins.io so we had to change the dns recored while we can still download from pkg.jenkins.io as we did, we cannot anymore push packages through that domain

### DIFF
--- a/env/release.mk
+++ b/env/release.mk
@@ -6,7 +6,7 @@
 export JENKINS_URL=https://cloudbees.ci.cloudbees.com/
 
 # the host to publish bits to
-export PKGSERVER=mirrorbrain@pkg.jenkins.io
+export PKGSERVER=mirrorbrain@52.202.51.185
 export SSH_OPTS=-p 22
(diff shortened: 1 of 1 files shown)


env/release.mk
```diff
@@ -6,7 +6,7 @@
 export JENKINS_URL=https://cloudbees.ci.cloudbees.com/
 
 # the host to publish bits to
-export PKGSERVER=mirrorbrain@pkg.jenkins.io
+export PKGSERVER=mirrorbrain@52.202.51.185
```
timja, line 9: why not create an ssh dns record?

olblak: like pkg.origin.jenkins.io ?

olblak: This will allow us to request certificate using letsencrypt in the futur

timja: yeah

PR #182, opened 2020-05-18
Add retry to signtool

slide:


msi/build/build.ps1
```diff
@@ -93,7 +93,14 @@ Get-ChildItem .\bin\Release -Filter *.msi -Recurse |
             Write-Host "Signing installer"
             # always disable tracing here
             Set-PSDebug -Trace 0
-            signtool sign /v /f $env:PKCS12_FILE /p $env:SIGN_STOREPASS /t http://timestamp.verisign.com/scripts/timestamp.dll /d "Jenkins Automation Server ${JenkinsVersion}" /du "https://jenkins.io" $_.FullName
+            while($True) {
+                $p = Start-Process -Wait -PassThru -NoNewWindow -FilePath "signtool.exe" -ArgumentList "sign /v /f `"${env:PKCS12_FILE}`" /p ${env:SIGN_STOREPASS} /t http://timestamp.verisign.com/scripts/timestamp.dll /d `"Jenkins Automation Server ${JenkinsVersion}`" /du `"https://jenkins.io`" $_.FullName"
+                $p.WaitForExit()
+                # we will retry until we get a good exit code
+                if($p.ExitCode -eq 0) {
```
timja, line 100: Do we not want to give up at some point or is there a timeout higher up that will handle this?

Should we sleep in between attempts to give the server more time to recover?

slide: A timeout would probably be good, it would just result in a failure, so not a bid deal. My guess is its not the server that is having issues, but the client accessing the server. Those timestamp servers have insane uptimes and response rates.

msi/build/build.ps1
```diff
@@ -93,7 +93,24 @@ Get-ChildItem .\bin\Release -Filter *.msi -Recurse |
...
+                    break
+                } else {
+                    Start-Sleep -Seconds 10
+                    $retries--
+                }
+            }
+            
+            if($retries -lt 0) {
+                Write-Error "signtool did not complete successfully after 10 tries"
```
olblak, line 110: hardcoding 10 seems weird to me and imho we should use a counter value instead of decreasing 'retries' so we could use retries value here but at least it does the job.

daniel-beck: This style combines the variable declaration with the limit, I like it.

slide: @olblak When you say counter value, you just want to count up to a certain value?

olblak: Sorry if you now set retries to 5, you'll still have a message saying `signtool did not complete successfully after 10 tries`
While I like your style of variable declaration and limit, once you arrive at the write-error, you lost the initial value which is why I would prefer something like 
```
for (i=$retries; i >0 ; i--){
  //
} 
System.out.println("signtool did not complete successfully after " + retries + "tries")
```

PR #188, opened 2020-06-25
Update Windows 'latest' symlink after uploading new msi

MarkEWaite:
## Update Windows 'latest' symlink after uploading new msi

The 'latest' symlink allows the Download page on jenkins.io to have a constant URL that points to http://mirrors.jenkins.io/windows/latest .  It relies on the production build process to update the symlink to point to the most recent build.

Step 1 to fix https://issues.jenkins-ci.org/browse/WEBSITE-758 and https://github.com/jenkins-infra/jenkins.io/issues/3476 .

The script creates the necessary symlink locally, uploads it with rsync, then removes the local symlink.

It seems like this should be able to use a pattern similar to the technique used to update the symbolic link to the war file parent directory in http://mirrors.jenkins.io/war/ .  It is not exactly the same, since this symbolic link needs to point to the MSI file, not to the parent directory of the MSI file.

Unfortunately, I was unable to find the location where the symlink is created for the parent directory of the war file.  @olblak or @slide can you point me to that location so that I can compare the techniques?

### DIFF
--- a/msi/publish/publish.sh
+++ b/msi/publish/publish.sh
@@ -94,9 +94,11 @@ function uploadPackage(){
   # This is a safety measure just in case something was left there previously
   rm -rf latest
 
-  # Create a local symlink pointing to the VERSION directory
-  # Don't need VERSION directory locally, just the unresolved symlink
-  ln -s ${VERSION} latest
+  # Create a local symlink pointing to the MSI file in the VERSION directory.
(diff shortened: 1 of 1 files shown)


slide:
I'm not sure where that is done, if it's a symlink, it would be done on Linux somewhere not in the windows scripts.

PR #192, opened 2020-07-24
Adding gpg key details (including fingerprints) to the RPM/Debian pages.

WillSkates:
Hello there,

I wanted to verify the GPG keys used to sign RPM packages and I couldn't find the key fingerprint anywhere.

This PR exports the public key info into a file. The content of that file gets loaded and input into the generated HEADER.html and index.html files for the Debian and RPM distributions.

The following PR is a work in progress. I've tested locally using the following commands:

```bash
mkdir -p ../jpkg-work
rm -f ../jpkg-work/*.html
DEB_URL="http://example.com" bin/indexGenerator.py --distribution debian --targetDir "$PWD/../jpkg-work" --gpg-key-info-file "$PWD/../jpkg-work/gpg.key.info"
```

AFAIK it _should_ work however it needs additional testing in a build environment which I can't do ATM.

### DIFF
--- a/bin/indexGenerator.py
+++ b/bin/indexGenerator.py
@@ -159,9 +159,8 @@ class IndexGenerator:
         if (self.gpg_pub_key_info_file != "."):
             gpg_pub_key = Path(self.gpg_pub_key_info_file)
             if (gpg_pub_key.is_file()): 
-                gpg_pub_key = open(self.gpg_pub_key_info_file, "r")
-                pub_key_info = gpg_pub_key.read()
-                gpg_pub_key.close()
+                with open(self.gpg_pub_key_info_file, "r") as gpg_pub_key:
+                    pub_key_info = gpg_pub_key.read()
         
         return pub_key_info
 
--- a/bin/indexGenerator.py
+++ b/bin/indexGenerator.py
@@ -1,11 +1,11 @@
 #!/usr/bin/env python3
 
 from jinja2 import Environment, FileSystemLoader
+from pathlib import Path
 
 import getopt
 import os
 import sys
-
 
 def basename(path):
     return os.path.basename(path)
@@ -68,12 +68,14 @@ class IndexGenerator:
         self.organization = os.getenv('ORGANIZATION', 'jenkins.io')
         self.product_name = os.getenv('PRODUCTNAME', 'Jenkins')
         self.distribution = os.getenv('OS_FAMILY', 'debian')
+        self.gpg_pub_key_info_file = os.getenv('GPGPUBKEYINFO', '.')
         self.target_directory = './target/' + self.distribution
+
         try:
             opts, args = getopt.getopt(
                 argv,
                 "hd:o:",
-                ["targetDir=", "distribution="]
+                ["targetDir=", "distribution=", "gpg-key-info-file="]
             )
         except getopt.GetoptError:
             print(self.HELP_MESSAGE)
@@ -85,6 +87,8 @@ class IndexGenerator:
             elif opt in ("-d", "--distribution"):
                 self.distribution = arg
                 self.target_directory = './target/' + self.distribution
+            elif opt in ("-g", "--gpg-key-info-file"):
+                self.gpg_pub_key_info_file = arg
             elif opt in ("-o", "--targetDir"):
                 self.target_directory = arg
                 self.targetFile = self.target_directory + "/HEADER.html"
@@ -112,6 +116,7 @@ class IndexGenerator:
         print('Repository footer generated: ' + self.footer)
         print('Root header generated: ' + self.root_header)
         print('Root footer generated: ' + self.root_footer)
+        print("GPG Key Info File: " + self.gpg_pub_key_info_file)
 
     def generate_root_header(self):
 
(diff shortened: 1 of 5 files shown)


olblak:
Thanks for opening this PR, everything that increases visibility is more than welcome.

While I still have to manually test this PR, I just want to add that the information you are looking for is kind of already available.

The public GPG key is published for each distribution, specifically for RedHat you can download it https://pkg.jenkins.io/redhat/jenkins.io.key then you can run `gpg --import-options show-only --import jenkins.io.key`

But not having to dig would be useful

olblak:
image

bin/indexGenerator.py
```diff
@@ -148,6 +153,18 @@ def generate_footer(self):
...
 
+    def fetch_pubkeyinfo(self):
+        pub_key_info = ""
+
+        if (self.gpg_pub_key_info_file != "."):
+            gpg_pub_key = Path(self.gpg_pub_key_info_file)
+            if (gpg_pub_key.is_file()): 
+                gpg_pub_key = open(self.gpg_pub_key_info_file, "r")
+                pub_key_info = gpg_pub_key.read()
+                gpg_pub_key.close()
```
MarkEWaite, line 164: Allow python to handle closing the file?
```suggestion
                with open(self.gpg_pub_key_info_file, "r") as gpg_pub_key:
                    pub_key_info = gpg_pub_key.read()
```

deb/publish/publish.sh
```diff
@@ -27,10 +27,12 @@ function generateSite(){
   cp -R "$bin/contents/." "$D/contents"
   
   gpg --export -a --output "$D/contents/${ORGANIZATION}.key" "${GPG_KEYNAME}"
+  echo "$(gpg --import-options show-only --import $D/${ORGANIZATION}.key)" > "$D/${ORGANIZATION}.key.info"
   
   "$BASE/bin/indexGenerator.py" \
     --distribution debian \
-    --targetDir "$D/html"
+    --targetDir "$D/html" \
+    --gpg-key-info-file "${D}/${ORGANIZATION}.key.info"
```
MarkEWaite, line 35: Reduce diffs?
```suggestion
    --gpg-key-info-file "${D}/${ORGANIZATION}.key.info" \
    --targetDir "$D/html"
```

rpm/publish/publish.sh
```diff
@@ -20,11 +20,13 @@ function clean(){
...
+  gpg --export -a --output "$D/${ORGANIZATION}.key" "${GPG_KEYNAME}"
+  echo "$(gpg --import-options show-only --import $D/${ORGANIZATION}.key)" > "$D/${ORGANIZATION}.key.info"
+  
   "$BASE/bin/indexGenerator.py" \
     --distribution redhat \
-    --targetDir "${D}"
-  
-  gpg --export -a --output "$D/${ORGANIZATION}.key" "${GPG_KEYNAME}"
+    --targetDir "${D}" \
+    --gpg-key-info-file "${D}/${ORGANIZATION}.key.info"
```
MarkEWaite, line 29: Reduce diffs?
```suggestion
    --gpg-key-info-file "${D}/${ORGANIZATION}.key.info" \
    --targetDir "${D}"
```

rpm/publish/publish.sh
```diff
@@ -20,11 +20,13 @@ function clean(){
 }
 
 function generateSite(){
+  gpg --export -a --output "$D/${ORGANIZATION}.key" "${GPG_KEYNAME}"
+  echo "$(gpg --import-options show-only --import $D/${ORGANIZATION}.key)" > "$D/${ORGANIZATION}.key.info"
```
MarkEWaite, line 24: This seems to be writing to the target directory that will be used by indexGenerator.py

Doesn't indexGenerator.py remove or replace existing files in the target directory?

WillSkates: Don't think so no. At least if it's in there I can't see it. All I find is open calls which open a particular output file for writing.

olblak: It copies the files to $D which is a temporary working directory used as the source directory that will be used to upload to both pkg.jenkins.io and mirror.jenkins.io. This means that next to ${ORGANIZATION}.key, we'll also have ${ORGANIZATION}.key.info.

deb/publish/publish.sh
WillSkates: Done.

rpm/publish/publish.sh
WillSkates: Done.

WillSkates:
Apologies @MarkEWaite I made your requested changes. I forgot to click "Ok" for the comments telling you so.

olblak:
I'll merge this PR after the weekly release is done

timja:
why not just include it?

timja:
unless you're planning to re-run it after release?

olblak:
> unless you're planning to re-run it after release?

Yes I am, it's just that I don't have a lot of availabilities at the moment so if for some reason it breaks the packaging job, then someone will have to fix it for today's release, but if we first generate the packages and then we merge this PR it gives us more time and doesn't impact end users.

To be honest, I am confident with this PR but I don't want to take any risks at the moment :)

olblak:
While it worked perfectly for redhat, It failed for Debian 
```
+ gpg --export -a --output /tmp/1230/contents/jenkins.io.key 62A9756BFD780C377CF24BA8FCEF32E745F2C3D5
++ gpg --import-options show-only --import /tmp/1230/jenkins.io.key
gpg: can't open '/tmp/1230/jenkins.io.key': No such file or directory
gpg: Total number processed: 0

```

https://pkg.jenkins.io/redhat/
https://pkg.jenkins.io/debian/

WillSkates:
https://github.com/jenkinsci/packaging/pull/194 <-- Possible fix.

PR #193, opened 2020-07-29
RELEASELINE is used in the MSI package name

olblak:
This [job](https://release.ci.jenkins.io/blue/organizations/jenkins/core%2Fpackage/detail/stable-2.235/4/artifacts) is failing because we are referencing a MSI package that doesn't contain the RELEASELINE. We weren't affected until now because weekly releases uses an empty releaseline


olblak:
I am going to merge it immediately as it isn't a big change

PR #194, opened 2020-08-04
Adding a (fix?) for debian output error.

WillSkates:
A fix for https://github.com/jenkinsci/packaging/pull/192#issuecomment-668600104

GPG keys are missing on https://pkg.jenkins.io/debian.


olblak:
@WillSkates Thanks for the PR, it's now working perfectly http://archives.jenkins-ci.org/debian/

PR #196, opened 2020-09-18
Fix MSI version

slide:
Fixes MSI version for versions with a minor version > 255. Also, fixes missing slash in directory and removes some debug output from the build.


timja (approved):

oleg-nenashev (approved):
OK?

MarkEWaite (approved):
Approving though I am not really competent to judge WiX script or PowerShell script

slide:
Just for clarification, this will do the following:

2.256 -> 2.255.2560
2.256.1 -> 2.255.2561
2.256.2 -> 2.255.2562
etc

Hopefully we don't have more than 9 LTS releases for a single baseline...

PR #201, opened 2020-11-25
Closing the crackes

sri-ka1ki:
Still alot of code to cleanup


MarkEWaite:
@sri-ka1ki the INFRA-910-warning branch has been merged into the master branch.  There is no need to merge from the master branch to the INFRA-910-warning branch.  Rather, we should consider removing the INFRA-910-warning branch since it is now a subset of the master branch

PR #203, opened 2020-12-21
jenkins.init: Use pgrep to match jenkins.war instead of all java processes.

parazyd:
This should improve some cornercases where other java processes might be
running as the same $JENKINS_USER.

### DIFF
--- a/deb/build/debian/jenkins.init
+++ b/deb/build/debian/jenkins.init
@@ -165,7 +165,7 @@ do_start()
 #
 get_running()
 {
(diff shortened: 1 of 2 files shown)


MarkEWaite (approved):
426f6a9e5a2b009cd7632e07b7ae71b01a6f04f4 was merged to the master branch in #118 .  We can ignore it in this pull request because the change already exists on the master branch.  I assume that is a GitHub artifact because the master branch has not been merged to this pull request.  That change was already delivered in Jenkins 2.263.1 as far as I can tell.

The `procps` Debian package is already required by the Jenkins package and provides the `pgrep` command used in this pull request.

timja (approved):

PR #209, opened 2021-08-11
Remove premature merge of weekly packaging change

MarkEWaite:
## Remove premature merge of weekly packaging change

The weekly packaging change is first available in Jenkins 2.306.  The next LTS release will be 2.302.1.  It won't have the weekly packaging change.  This change was merged into the stable-2.302 branch before the matching code to support it was available in the LTS branch.  Remove the commits from that premature merge.  Can be merged in 3 months for the next LTS.

- Revert "[JENKINS-66347] Keystore not working (#208)"
- Revert "Trade our global dependency on Akuma for an RPM-specific dependency on daemonize (#207)"

@timja @basil this is correcting a mistake I made on the stable-2.302 branch of the packaging repository.


basil (approved):

timja:
What matching code is required? I.e why not include this? Although I assume no rush is required for this

MarkEWaite:
> What matching code is required? I.e why not include this? Although I assume no rush is required for this

When I merged this into the stable-2.302 branch I effectively backported the change from 2.306 without sufficient time for the community to evaluate the change and without any comment from others on including the change.  I think that was a mistake.

I expect a number of users will be surprised by a new dependency on a program that is in an optional repository.  I'd rather wait one more LTS cycle before we include this.  That might even be enough time to make the transition from init based to systemd.

timja:
Sure that makes more sense, but the description in the pull request doesn't match that 😄

PR #214, opened 2021-08-29
fix:Change to use 127.0.0.1 instead of localhost in browser launch at end of install 

slide:
Change to use 127.0.0.1 instead of localhost, this fixes https://issues.jenkins.io/browse/JENKINS-41101

### DIFF
--- a/msi/build/jenkins.wxs
+++ b/msi/build/jenkins.wxs
@@ -460,7 +460,7 @@
       <Publish Dialog="MaintenanceTypeDlg" Control="RemoveButton" Event="NewDialog" Value="VerifyReadyDlg">1</Publish>
       <Publish Dialog="MaintenanceTypeDlg" Control="Back" Event="NewDialog" Value="MaintenanceWelcomeDlg">1</Publish>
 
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):

daniel-beck:
Isn't there a dependency on IPv4 for 127.0.0.1? Is that always present on Windows?

I think we implemented the opposite change somewhere else not too long ago.

slide:
> Isn't there a dependency on IPv4 for 127.0.0.1? Is that always present on Windows?

No, IPv4 is not always present (it can be disabled)

> 
> I think we implemented the opposite change somewhere else not too long ago.

Do you remember where this was? I'd like to look at the discussion if possible.

daniel-beck:
https://github.com/jenkinsci/maven-hpi-plugin/pull/169 might be what I thought was the change "not too long ago" (yay memory), but I cannot find what motivated that position specifically.

https://github.com/jenkinsci/jenkins-test-harness/pull/9 was before, but I probably just checked this for consistency/completeness – but FWIW another point for the `localhost` in acrive use, even though all of these are in non-production environments.

I mean, if I'm wrong here it wouldn't be the first time 😄 What is this change supposed to accomplish? I read the ticket but don't understand it.

slide:
Honestly, now that I look at the issue again, I think its bogus. Sorry for the cruft.

PR #218, opened 2021-10-10
Avoid killing jenkins on logrotate

bmwiedemann:
copied from the `rpm` version

In my tests, SIGALRM terminated jenkins and caused long-running jobs to fail


basil:
Seems reasonable at face value. Has this actually been tested in a production use case?

bmwiedemann:
Yes, we are using this on our two production instances at SUSE.

basil (approved):

PR #223, opened 2021-11-08
Use keyring instead of apt-key to register Debian signing key on stable-2.303

MarkEWaite:
## Use keyring instead of apt-key for Debian install on stable-2.303

See https://github.com/jenkins-infra/jenkins.io/pull/4675 for the rationale and the links to information like:

* https://unix.stackexchange.com/a/463140/498490
* https://wiki.debian.org/DebianRepository/UseThirdParty

Include this change on the stable-2.303 branch because the next LTS branch will likely be constructed from stable-2.303.

### DIFF
--- a/templates/header.debian.html
+++ b/templates/header.debian.html
@@ -13,14 +13,17 @@
 
   <pre class="text-white bg-dark">
     <code>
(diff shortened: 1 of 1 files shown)


timja (changes requested):
stable branches are created from master, no need for this

PR #238, opened 2022-01-17
Add MSI build to `Jenkinsfile`

basil:
I have no idea if this will work or not.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,7 @@ properties([
   buildDiscarder(logRotator(numToKeepStr: '15', artifactNumToKeepStr: '5')),
   disableConcurrentBuilds(abortPrevious: true)
 ])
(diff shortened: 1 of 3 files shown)


basil:
@slide I have not been following the infrastructure team's work too closely, but based on a quick readthrough of jenkins-infra/helpdesk#2746 it seems that they switched to a new system of VM (as opposed to container) based Windows agents for the official release process in https://github.com/jenkins-infra/release

I think the next steps are to reverse engineer whatever they implemented for the official release process and get it up and running in the `Jenkinsfile` for this repository and on `ci.jenkins.io`, working with the infrastructure team as necessary if the same VM-based build environment is not already available on `ci.jenkins.io`. That is the essence of what I was trying to do in this PR, but it is now out of date (both physically in the sense of merge conflicts and logically in the sense that it's based on the old container deployment), so I will close this PR.

PR #241, opened 2022-01-18
Test on more Ubuntu versions

basil:


molecule/default/molecule.yml
```diff
@@ -4,6 +4,47 @@ dependency:
 driver:
   name: docker
 platforms:
+  - name: ubuntu-14-04
```
timja, line 7: Do we want to test against 14 and 16 given they are out of LTS support?

MarkEWaite: I don't think we should test any Ubuntu version older than 18.04.

basil: No strong preference. I can remove them if you want.

basil: I removed Ubuntu 14.04 and 16.04 and added Debina 10 and 11.

MarkEWaite: That's perfect!  Debian 10 (oldstable) and Debian 11 (stable) are the right additions.

Jenkinsfile
```diff
@@ -33,14 +33,15 @@ podTemplate(yaml: readTrusted('KubernetesPod.yaml'), workingDir: '/home/jenkins/
   }
 }
 
-nodeWithTimeout('azure && docker') {
+nodeWithTimeout('docker') {
```
dduportal, line 36: quick note: if you want more resources on the VM, you can use `docker-highmem` (16 vCPUS / 64 Gb instead of the 4vCPUS / 16 Gb for `docker`).

timja (approved):

Jenkinsfile
basil: Thank you for the advice. I might explore that if I want to try running the tests in parallel at some point.

PR #243, opened 2022-01-18
Add Rocky Linux 8 and Alma Linux 8

basil:


molecule/default/molecule.yml
```diff
@@ -53,6 +54,14 @@ platforms:
       - ${MOLECULE_PROJECT_DIRECTORY}/target:/var/tmp/target
       - /sys/fs/cgroup:/sys/fs/cgroup:ro
     privileged: true
+  # rpm
+  - name: almalinux-8
```
timja, line 58: aren't these essentially the same thing? do we get enough benefit for the 'cost' of running?

(non blocking)

timja (approved):

molecule/default/molecule.yml
basil: I don't have a dog in the Rocky vs Alma race.

PR #245, opened 2022-01-18
Add CentOS 8

basil:


MarkEWaite:
I would not include CentOS 8 in the testing because it is an abandoned operating system.  Red Hat has declared a different direction for CentOS with the announcement of [CentOS Stream](https://blog.centos.org/2020/12/future-is-centos-stream/).  I think the testing of Red Hat Enterprise Linux 8, Rocky Linux 8, Alma Linux 8, or Oracle Linux 8 would be a better choice than testing CentOS 8.

Requests have been received for better support of Amazon Linux 2.  It is an RPM based distribution that was forked from Red Hat Enterprise Linux some time ago (at least as far as I can tell).

Amazon Linux 1 ended support on Dec 31, 2020.  I don't think we should test with it.

basil:
Can you be more explicit about you the action item in your review Mark? I already added Rocky Linux 8 and Alma Linux 8 in #243. RHEL is not available in `dokken-images`. So are you suggesting this PR be closed? Please just say so then.

MarkEWaite:
Sure.  I recommend this pull request be closed.  We've removed CentOS 8 from the Jenkins Docker images and added almalinux and UBI images.  I don't think we should automate the testing of CentOS 8.

PR #246, opened 2022-01-18
Add CentOS Stream 8 and Oracle Linux 8

basil:


MarkEWaite (approved):

basil:
The Alma Linux mirror is down so I need to do another testing cycle on this PR, so might as well include CentOS Stream 8 in the diff. BTW @MarkEWaite the move to `systemd` might be more urgent than we thought, as `daemonize` is missing in Fedora EPEL 9:

https://src.fedoraproject.org/rpms/daemonize

PR #248, opened 2022-01-19
Drop daemonize dependency

basil:
Users have expressed displeasure with the dependency on `daemonize(1)`. When I first introduced this dependency, I only tested on Fedora, naively assuming that this package is available everywhere. Needless to say, it isn't, and setting up the corresponding EPEL repository has been a headache for users. Moreover, it is not available at all in EPEL 9:

https://src.fedoraproject.org/rpms/daemonize

This PR removes the dependency by calling the `daemon` built-in shell function. That function doesn't background the process, but I do so with an `&`.

### Testing Done

The automated tests for this repository install the package on all supported RPM distributions. I ran the tests locally in Docker and they passed. Note the changes to the test cases: EPEL is no longer installed, thus we are testing that the package correctly installs without EPEL.


basil:
Having a problem with the Java process not being terminated after stopping the service in manual testing... back to draft.

basil:
I see the problem... when backgrounding the process with `&` we need to wait for a little for it to start before trying to record its PID. No big deal, just a small matter of programming.

basil (commented):
Manual testing now passes. I verified that starting Jenkins results in the Java process running and the correct value in the PID file. I verified that stopping Jenkins stops the Java process using the PID stored in the PID file.

basil:
I improved automated test coverage so that it would have caught the mistake from my earlier revision. Automated tests now:

- Install the package.
- Ensure Java is not running.
- Start the service.
- Ensure Java is running.
- Ensure HTTP is serving a 403 login screen.
- Restart the service.
- Ensure Java is running.
- Stop the service.
- Ensure Java is not running.

These automated tests have passed for all platforms:

- AlmaLinux 8
- Amazon Linux 2
- CentOS Stream 8
- Debian 10
- Debian 11
- Fedora 34
- Fedora 35
- openSUSE Leap 15
- Oracle Linux 8
- Rocky Linux 8
- Ubuntu 18.04
- Ubuntu 20.04
- Ubuntu 21.04
- Ubuntu 21.10
- Ubuntu 22.04

timja (approved):

basil:
The new tests are mostly passing, but not consistently so... there's #249, which we've seen on the main branch before this PR, and a new flake in a new test which I'm not sure is due to this PR or a pre-existing issue. To eliminate the ambiguity, I'm planning to first fix #249, then extract the new tests out of this PR into a new PR and run them a few times against trunk and merge the result if passing. Then I plan to rebase this PR on top of trunk and run the PR build a few times. This way we can be sure of the cause of any flakiness in tests.

basil:
I have split this PR into two PRs: #253, which improves test coverage and fixes #249, and this PR (based on top of #253), which then adds the changes to drop daemonize. All tests are passing both locally and in CI, so I have marked this as ready to review. My preference would be to squash-merge #253 first, then to squash merge this PR (to maintain a clean Git history).

basil:
With all prerequisite PRs merged, 2 consecutive passing test runs in CI, and multiple passing test runs locally across the entire platform matrix, this PR is ready to merge from my perspective.

@MarkEWaite Did you have a preference as to when we merge this? My preference would be for this to be merged now, to be delivered in the 2.333 weekly, and then (assuming no issues are discovered) to be delivered in the 2.319.3 LTS release. Users seem to hate the `daemonize` dependency, so I think they would be relieved to see it gone. However, I could see an argument for waiting until after 2.319.3 is delivered to merge this PR to mitigate risk. I think the risk is low since automated tests are covering this code path for all supported platform, but I have no strong preference one way or another. Happy to go with the group consensus here.

MarkEWaite:
I'd prefer to wait until after the 2.319.3 release is complete.  We can include the change in a weekly after 2.319.3 is delivered and include it in the 2.332.1 changelog and upgrade guide.  I'm happy to provide a blog post or other communication alerting people to the change.

I think this one might be good in combination with the proposed change to the Docker lifecycle management.  Both are "Unix-related" changes and they might fit well together in a blog post.

timja:
It’s using a different branch for the next LTS release anyway so you can merge whenever

timja (approved):

basil:
> It’s using a different branch for the next LTS release anyway so you can merge whenever

Oh right, thanks for pointing that out. So I will merge this now, to be delivered in the 2.333 weekly but _not_ the 2.319.3 LTS release.

Issue #249, opened 2022-01-19
openSUSE Leap testing is flaky


basil:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
  Paste the output here
  ```

</details>

### What Operating System are you using (both controller, and any agents involved in the problem)?

openSUSE Leap 15.3 x86_64

### Reproduction steps

Run `molecule test` with no changes as described in the `README.`

### Expected Results

The tests pass.

### Actual Results

Observed in https://ci.jenkins.io/job/Packaging/job/packaging/job/master/16/consoleFull and https://ci.jenkins.io/job/Packaging/job/packaging/job/PR-248/4/consoleFull

```
fatal: [opensuse-leap-15]: FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/usr/bin/python3.6"}, "changed": false, "msg": "Unable to start service jenkinstest: Job for jenkinstest.service failed because the control process exited with error code.\nSee \"systemctl status jenkinstest.service\" and \"journalctl -xe\" for details.\n"}
```

### Anything else?

I can't reproduce the failure locally, and it often passes in re-runs on CI.

basil:
`journalctl -u` output for the failure:

```
-- Logs begin at Thu 2022-01-20 18:26:26 UTC, end at Thu 2022-01-20 18:35:18 UTC. --
Jan 20 18:35:09 opensuse-leap-15 systemd[1]: Starting LSB: Jenkins Automation Server (Test)...
Jan 20 18:35:09 opensuse-leap-15 jenkinstest[1686]: + JENKINS_WAR=/usr/lib/jenkinstest/jenkinstest.war
Jan 20 18:35:09 opensuse-leap-15 jenkinstest[1686]: + test -r /usr/lib/jenkinstest/jenkinstest.war
Jan 20 18:35:09 opensuse-leap-15 jenkinstest[1686]: + JENKINS_CONFIG=/etc/sysconfig/jenkinstest
Jan 20 18:35:09 opensuse-leap-15 jenkinstest[1686]: + test -r /etc/sysconfig/jenkinstest
Jan 20 18:35:09 opensuse-leap-15 jenkinstest[1686]: + JENKINS_PID_FILE=/var/run/jenkinstest.pid
Jan 20 18:35:09 opensuse-leap-15 jenkinstest[1686]: + JENKINS_USER=jenkinstest
[...]

PR #250, opened 2022-01-19
Debugging #249 by telegram

basil:


basil:
Fix has been proposed in #253.

PR #261, opened 2022-01-27
Drop dependency on `daemon(1)` in `.deb` packages

basil:
This PR removes about 100 lines from the Debian initialization script by dropping the dependency on `daemon(1)`, `su(1)`, `pgrep(1)`, and `killall(1)` (provided by the `daemon`, `procps`, and `psmisc` packages) in favor of `start-stop-daemon(8)` and `init-d-script(5)` (provided by the `lsb-base` and `sysvinit-utils` packages). This significantly cleans up both the code (by leaning on `/lib/init/init-d-script` to do the heavy lifting) and the process tree (by eliminating the `daemon` and `su` processes).

I tested a number of scenarios manually:

- Ensure that startup fails when `RUN_STANDALONE=false`
- Ensure that startup fails with a proper error message when the port is in use
- Ensure that a custom `ulimit` is set
- Ensure that a custom `umask` is set
- Ensure that the process's environment, current working directory, and arguments match the ones from before this PR
- Ensure that `apt purge` removes the home directory


deb/build/debian/control
```diff
@@ -1,12 +1,12 @@
 Source: @@ARTIFACTNAME@@
 Section: devel
-Priority: extra
+Priority: optional
```
basil, line 3: Per [this page](https://wiki.debian.org/Packaging/Intro):

>  As of [Debian Policy 4.5.1](https://www.debian.org/doc/debian-policy/ch-archive.html#s-priorities) (or sooner) package priority `extra` is deprecated.

deb/build/debian/control
```diff
@@ -1,12 +1,12 @@
...
 Maintainer: @@AUTHOR@@
-Build-Depends: debhelper (>= 9)
-Standards-Version: 3.9.4
+Build-Depends: debhelper (>= 10)
+Standards-Version: 3.9.8
```
basil, line 6: Moving from [August 2012 to April 2016](https://www.debian.org/doc/packaging-manuals/upgrading-checklist.txt).

deb/build/debian/control
```diff
@@ -1,12 +1,12 @@
...
 Homepage: @@HOMEPAGE@@
 
 Package: @@ARTIFACTNAME@@
 Architecture: all
-Depends: ${misc:Depends}, daemon, adduser, procps, psmisc, net-tools
+Depends: ${misc:Depends}, adduser, lsb-base (>= 3.2-14), net-tools, sysvinit-utils
```
basil, line 11: `lsb-base` >= 3.2-14 is needed for `status_of_proc`, and `sysvinit-utils` is needed for `/lib/init/init-d-script` (see below).

deb/build/debian/jenkins.init
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env /lib/init/init-d-script
```
basil, line 1: See [`init-d-script(5)`](https://manpages.debian.org/buster/sysvinit-utils/init-d-script.5.en.html) for more information about this library.

deb/build/debian/jenkins.init
```diff
@@ -13,91 +13,54 @@
 # Description:       Controls @@SUMMARY@@
 ### END INIT INFO
 
-PATH=/bin:/usr/bin:/sbin:/usr/sbin
-
+DAEMON=none
```
basil, line 16: We do our own handling of the Java command, so set this to `none` to opt out of the handling offered by `init-d-script(5)`).

deb/build/debian/jenkins.init
```diff
@@ -13,91 +13,54 @@
...
-	local assigned=$2
-	local default=$3
-	local assigned_address=$4
-	local default_address=$5
-	local port
+	service=$1
+	assigned=$2
+	default=$3
+	assigned_address=$4
+	default_address=$5
```
basil, line 63: `local` is `bash`-specific. We don't depend on the `bash` package, so we shouldn't be using `bash`isms.

deb/build/debian/jenkins.postrm
```diff
@@ -4,8 +4,16 @@ set -e
...
+	if [ -x "$(command -v deluser)" ]; then
+		deluser --quiet --system @@ARTIFACTNAME@@ >/dev/null || true
+	else
+		userdel @@ARTIFACTNAME@@ || echo "failed to remove @@ARTIFACTNAME@@ system user" >&2
+	fi
+	if [ -x "$(command -v delgroup)" ]; then
+		delgroup --quiet --system --only-if-empty @@ARTIFACTNAME@@ >/dev/null || true
+	else
+		groupdel @@ARTIFACTNAME@@ || echo "failed to remove @@ARTIFACTNAME@@ system group" >&2
+	fi
```
basil, line 16: Following the advice from [this page](https://wiki.debian.org/AccountHandlingInMaintainerScripts).

timja (approved):

haegar:
Just a minor heads-up:

Before this PR jenkins was using the $PATH of the jenkins user, which included /usr/local/bin by default - now it is using the more restricted $PATH from the init.d script, which does not contain it.

This made some of our builds fail, as commands installed in /usr/local/bin were no longer found when executed without full path.

I managed to work around this by adding a custom PATH=... line to /etc/default/jenkins

basil:
@haegar [Running jobs on the controller is not recommended to begin with.](https://www.jenkins.io/doc/book/security/controller-isolation/)

Even still, jobs should not be relying on the `PATH` set by the initialization script. That is an implementation detail that is subject to change at any time.

If you need to set a custom `PATH` (or any environment variable) for jobs, the recommendation is to do so in:

-  **Manage Jenkins, Configure System, Global Properties, Environment Variables** (if the variable applies to all nodes)
- **Manage Jenkins, Manage Nodes and Clouds, Configure, Node Properties, Environment Variables** (if the variable is node-specific)

Both of the above can be configured with Jenkins Configuration as Code (JCasC).

bra-fsn:
> @haegar [Running jobs on the controller is not recommended to begin with.](https://www.jenkins.io/doc/book/security/controller-isolation/)
> 
> Even still, jobs should not be relying on the `PATH` set by the initialization script. That is an implementation detail that is subject to change at any time.
> 
> If you need to set a custom `PATH` (or any environment variable) for jobs, the recommendation is to do so in:
> 
>     * **Manage Jenkins, Configure System, Global Properties, Environment Variables** (if the variable applies to all nodes)
> 
>     * **Manage Jenkins, Manage Nodes and Clouds, Configure, Node Properties, Environment Variables** (if the variable is node-specific)
> 
> 
> Both of the above can be configured with Jenkins Configuration as Code (JCasC).

I'm not sure what this solves, but I think it was a good feature to automatically inherit (for example) the `PATH` from the owning user. That might come from anywhere, even set up dynamically (LDAP and similar configuration places anybody?) and with this change you made those break without any nice ways to keep that functionality.
BTW, there are plugins for wiping those envvars, now they must be sad plugins. :(
Still, I don't understand what's the benefit of removing this feature. And why this isn't mentioned in the changelog?
It's quite a breaking change for many users, and all of them won't understand what happened and will spend hours at worst to figure it out.
I think this would've deserved more caution and documentation (yet agreement from the users).

timja:
> I think this would've deserved more caution and documentation (yet agreement from the users).

It was a side effect, and an edge-case.

It's mentioned in the changelog as:
>  Remove requirement for daemon in deb packages

bra-fsn:
> Remove requirement for daemon in deb packages

:100:

kugel-:
This change seems to have broken (unprivileged) LXC containers on our setup (they are started by Jenkins jobs). Starting the container via SSH (as the `jenkins` user) works. `start-stop-daemon` seems to make a difference, perhaps related to cgroups or unprivilged containers.

I reverted to Jenkins 2.331 for now, back in the green.

timja:
2.335 (releasing today) will bring systemd, maybe that will help?

kugel-:
You mean a native unit? I'll give that a try, thanks for the info.

timja:
yes

basil:
@kugel- [Running jobs on the controller is not recommended to begin with.](https://www.jenkins.io/doc/book/security/controller-isolation/)

Even still, jobs running on the controller should not be relying on the controller Java process having any particular set of privileges other than those needed to run Jenkins itself. That is an implementation detail that is subject to change at any time.

From [the Jenkins documentation](https://www.jenkins.io/doc/book/security/controller-isolation/#not-building-on-the-built-in-node):

> If you do not have any other computers to run agents on, you can also run an agent process as a different operating system user on the same system.

If you need to run jobs on the controller with a specific set of privileges, the recommendation is to run an agent process as a different operating system user on the same system with the desired set of privileges.

kugel-:
Thanks for the hint with the local agent user. So how do you connect that user to the controller? ssh to localhost, websockets or just tcp over localhost? That all seems overkill. The documentation is lacking a recommendation for such a setup.

Would a systemd unit with `User=jenkings-agent` and `ExecStart=java … agent.jar` suffice?

Also, a related question: Without explicit `node { … }` step, the entire pipeline runs on the controller. What happens if running jobs on the controller is disabled? Can I assign a different "fallback" agent? Many of our jobs don't use explicit `node` steps so I'm concerned they'd break quickly.

basil:
@kugel- These questions are more general and not specific to this PR. Please use the [forum](https://community.jenkins.io/), the [users mailing list](https://www.jenkins.io/mailing-lists/), or the various [chat rooms](https://www.jenkins.io/chat/) for further discussion.

Issue #217, opened 2021-09-29
Add --add-opens to installations running on java 17


timja:
See https://github.com/jenkinsci/docker/pull/1211 where it was added to docker images

PR #266, opened 2022-01-28
Support for `systemd`

### LINKED ISSUE #216: Systemd support
There is a PR here where inspiration can be taken from: https://github.com/jenkinsci/packaging/pull/93

basil:
Downstream of jenkins-infra/docker-packaging#29. Beginnings of a common `systemd.service(5)` unit configuration for the `deb`, `rpm`, and `suse` targets.

Should fix most if not all of the following bugs:

- https://issues.jenkins.io/browse/JENKINS-31656
- https://issues.jenkins.io/browse/JENKINS-41218
- https://issues.jenkins.io/browse/JENKINS-46020
- https://issues.jenkins.io/browse/JENKINS-51888
- https://issues.jenkins.io/browse/JENKINS-54558
- https://issues.jenkins.io/browse/JENKINS-55053
- https://issues.jenkins.io/browse/JENKINS-55696
- https://issues.jenkins.io/browse/JENKINS-56219
- https://issues.jenkins.io/browse/JENKINS-57015
- https://issues.jenkins.io/browse/JENKINS-60763
- https://issues.jenkins.io/browse/JENKINS-63803
- https://issues.jenkins.io/browse/JENKINS-66277
- https://issues.jenkins.io/browse/JENKINS-67404
- https://issues.jenkins.io/browse/JENKINS-67487

Fixes #216


basil:
[First successful CI test run](https://ci.jenkins.io/job/Packaging/job/packaging/job/PR-266/8/consoleFull) with `systemd` on all three targets: `deb`/`rpm`/`suse`.

We now have a distribution-agnostic `systemd` service unit (`jenkins.service`), as well as a distribution-agnostic launcher script (`/usr/bin/jenkins`) that accepts a well-defined set of environment variables (or no input at all, in which case it is the equivalent of `java -jar jenkins.war`). The `systemd` service unit defines sensible default options, and these options can be overridden with a systemd drop-in unit (which the tests do to customize `-Xmx`).

The main work that remains is the migration script to copy over old settings from `/etc/{default,sysconfig}` to the systemd drop-in unit, if the user has customized old defaults. I have the beginnings of such a script, but it needs more work. I haven't included it in the build yet or figured out how to invoke it, and it has a number of non-trivial TODOs remaining before I can even start testing.

basil:
I implemented a migration routine for all three targets and did some light testing. It seems to work fine. I'll continue to do more extensive testing, but at this point I think this change is ready for review.

basil:
Tested "Restart Jenkins when installation is complete and no jobs are running" with `ExitLifecycle`, worked as expected.

jtnord:
If you are moving the package of files why not move it to the correct location somewhere under `/opt/jenkins/`? (along with other files)

Other than that CloudBees no loger uses these scripts so the branding, name parameters etc could probably be cleaned up if it makes your life easier.

Other than that seems ok.

basil:
> If you are […] why not

Because I lack the energy to add both `systemd(1)` support and FHS compliance (lack of which is a preëxisting issue) in the scope of a single PR. Adding `systemd(1)` support was work enough. I am unwilling to increase the scope of this PR further.

> Other than that CloudBees no loger uses these scripts so the branding, name parameters etc could probably be cleaned up if it makes your life easier.

If I was interested in making my life easier, I would not have opened this PR.

> Other than that seems ok.

Thank you for the vote of confidence, James.

timja (approved):

deb/build/debian/jenkins.default
```diff
@@ -20,7 +20,7 @@ JENKINS_USER=$NAME
 JENKINS_GROUP=$NAME
 
 # location of the jenkins war file
-JENKINS_WAR=/usr/share/$NAME/$NAME.war
+JENKINS_WAR=/usr/share/java/$NAME.war
```
timja, line 23: is this supposed to be `/usr/share/jenkins`?

basil: Please read [this](https://wiki.debian.org/Java/Packaging) and [this](https://docs.fedoraproject.org/en-US/packaging-guidelines/Java/) before leaving such comments.

timja: If you expect people to read such things at least link them in the description, (which I checked before asking)

basil: > If you expect

I have no such expectation, but I do expect that someone asking "is this supposed to be X?" would provide their reasoning.

MarkEWaite:
I have two systems that I intend to test with the build from this pull request.  

One is a Red Hat Enterprise Linux 8 configuration that I just installed and configured with my test environment.  It is running Jenkins 2.331 currently with multiple environment variables and with a configuration change to use Java 11.  Agents are confirmed running on the configuration and jobs are being processed as expected.  I'll create a backup of the test environment then run the upgrade to explore the transition from System V init to systemd on Red Hat Enterprise Linux.

The second test system is a Debian 10 configuration running a subset of my test environment.  It's configuration is simpler than the RHEL 8 system, but it has been running longer.  It is configured behind an nginx reverse proxy.

basil:
> I have two systems that I intend to test with the build from this pull request.

Just be careful because the `.deb` files produced by the CI build use the test branding and are named `jenkinstest`, use a user named `jenkinstest`, a directory of `/var/lib/jenkinstest`, etc. You can't simply upgrade a production build to one of these.

basil:
Starting with [PR build #21](https://ci.jenkins.io/job/Packaging/job/packaging/job/PR-266/21/) these are using the regular branding and should be easier to test.

ekohl:
> Starting with [PR build #21](https://ci.jenkins.io/job/Packaging/job/packaging/job/PR-266/21/) these are using the regular branding and should be easier to test.

Looking at the built RPMs:
```console
$ rpm -q --scripts jenkins-2.332-1.2.noarch.rpm
warning: jenkins-2.332-1.2.noarch.rpm: Header V4 RSA/SHA512 Signature, key ID 41ed7c7f: NOKEY
preinstall scriptlet (using /bin/sh):
/usr/sbin/groupadd -r jenkins &>/dev/null || :
# SUSE version had -o here, but in Fedora -o isn't allowed without -u
/usr/sbin/useradd -g jenkins -s /bin/false -r -c "Jenkins Automation Server" \
	-d "/var/lib/jenkins" jenkins &>/dev/null || :
postinstall scriptlet (using /bin/sh):
/usr/share/jenkins/migrate "/etc/sysconfig/jenkins" || true

 
if [ $1 -eq 1 ] && [ -x "/usr/lib/systemd/systemd-update-helper" ]; then 
    # Initial installation 
    /usr/lib/systemd/systemd-update-helper install-system-units jenkins.service || : 
fi
preuninstall scriptlet (using /bin/sh):

 
if [ $1 -eq 0 ] && [ -x "/usr/lib/systemd/systemd-update-helper" ]; then 
    # Package removal, not upgrade 
    /usr/lib/systemd/systemd-update-helper remove-system-units jenkins.service || : 
fi
postuninstall scriptlet (using /bin/sh):

 
if [ $1 -ge 1 ] && [ -x "/usr/lib/systemd/systemd-update-helper" ]; then 
    # Package upgrade, not uninstall 
    /usr/lib/systemd/systemd-update-helper mark-restart-system-units jenkins.service || : 
fi
```

The `/usr/lib/systemd/systemd-update-helper` helper does not exist in EL7 so all actions will be noop.

You either need specific builds for each platform or have macros that support all OS versions.

basil:
> The `/usr/lib/systemd/systemd-update-helper` helper does not exist in EL7 so all actions will be noop.

EL7 is no longer supported by Red Hat as of August 30, 2021, right? Thus the Jenkins project does not support it, either.

basil:
As of [PR build 23](https://ci.jenkins.io/job/Packaging/job/packaging/job/PR-266/23/), startup completion notifications are enabled (see jenkinsci/jenkins#6228) and the RPMs support systemd v219 and later (see jenkinsci/packaging#266).

@ekohl @MarkEWaite It would be great if you could start testing the packages from that build.

ekohl:
Just for completeness, the scripts look good to me:
```console
$ rpm -q --scripts jenkins-2.333-1.2.noarch.rpm
warning: jenkins-2.333-1.2.noarch.rpm: Header V4 RSA/SHA512 Signature, key ID 41ed7c7f: NOKEY
preinstall scriptlet (using /bin/sh):
/usr/sbin/groupadd -r jenkins &>/dev/null || :
# SUSE version had -o here, but in Fedora -o isn't allowed without -u
/usr/sbin/useradd -g jenkins -s /bin/false -r -c "Jenkins Automation Server" \
	-d "/var/lib/jenkins" jenkins &>/dev/null || :
postinstall scriptlet (using /bin/sh):
/usr/share/jenkins/migrate "/etc/sysconfig/jenkins" || true

if [ $1 -eq 1 ] ; then 
        # Initial installation 
        systemctl preset jenkins.service >/dev/null 2>&1 || : 
fi
preuninstall scriptlet (using /bin/sh):

if [ $1 -eq 0 ] ; then 
        # Package removal, not upgrade 
        systemctl --no-reload disable jenkins.service > /dev/null 2>&1 || : 
        systemctl stop jenkins.service > /dev/null 2>&1 || : 
fi
postuninstall scriptlet (using /bin/sh):

systemctl daemon-reload >/dev/null 2>&1 || : 
if [ $1 -ge 1 ] ; then 
        # Package upgrade, not uninstall 
        systemctl try-restart jenkins.service >/dev/null 2>&1 || : 
fi
```

I'll try to test it an actual VM tomorrow but I can't promise anything.

basil:
Those testing HTTPS keystores will want to use the artifacts from [PR build 25](https://ci.jenkins.io/job/Packaging/job/packaging/job/PR-266/25/).

MarkEWaite:
> @ekohl @MarkEWaite It would be great if you could start testing the packages from that build.

I've installed build 25 on my Red hat Enterprise Linux 8 machine that was previously running 2.331.  Now running Jenkins 2.333-rc32035.1a_4204434750 and exploring.

rpm/build/SPECS/jenkins.spec
```diff
@@ -39,15 +41,15 @@ Authors:
...
+%__install -D -m0644 "%{SOURCE0}" "%{buildroot}%{_javadir}/%{name}.war"
 %__install -d "%{buildroot}%{workdir}"
 %__install -d "%{buildroot}%{workdir}/plugins"
 
 %__install -d "%{buildroot}/var/log/%{name}"
 %__install -d "%{buildroot}/var/cache/%{name}"
 
 %__install -D -m0755 "%{SOURCE1}" "%{buildroot}/etc/init.d/%{name}"
-%__sed -i 's,~~WAR~~,%{_prefix}/%{name}.war,g' "%{buildroot}/etc/init.d/%{name}"
+%__sed -i 's,~~WAR~~,%{_javadir}/%{name}.war,g' "%{buildroot}/etc/init.d/%{name}"
```
ekohl, line 52: Is the whole `init.d` part still relevant? Can't you just drop it?

systemd/jenkins.service
```diff
@@ -0,0 +1,142 @@
+[Unit]
+Description=Jenkins Continuous Integration Server
+Requires=network.target
+After=network.target
+
+#
+# Note: Do not customize this section. Instead, place your customizations in
+# the following drop-in file:
+#
+#     /etc/systemd/system/@@ARTIFACTNAME@@.service.d/override.conf
```
ekohl, line 10: Perhaps it should mention `systemctl edit @@ARTIFACTNAME.service@@` as a way to customize.

ekohl (commented):
I haven't tested this yet, but I wonder if the init.d and sysconfig files shouldn't simply be dropped for all Red Hat based distros, possibly Suse too. Since EL7 it only supports systemd (unlike Debian which still supports sysvinit as an option) so it won't be used anyway if there is a .service file.

And I can sympathize with wanting to use newer systemd features and finding out older platforms lack the support.

dhs-rec:
Tested the Debian package: Works.
However, I can't find anything in the systemd manpages about the (commented out) `SetCredential` and `LoadCredential` settings in `/lib/systemd/system/jenkins.service`.

timja:
> Tested the Debian package: Works. However, I can't find anything in the systemd manpages about the (commented out) `SetCredential` and `LoadCredential` settings in `/lib/systemd/system/jenkins.service`.

See https://github.com/jenkinsci/packaging/pull/266/commits/d73dc1ea2a0eba9f4fdc166a73edf6fce0f8da3e

Only very recent versions support that, it was reverted to use the more widely supported options

dhs-rec:
Ah, OK. Thanks a lot.

basil:
> I haven't tested this yet, but I wonder if the init.d and sysconfig files shouldn't simply be dropped for all Red Hat based distros, possibly Suse too. Since EL7 it only supports systemd (unlike Debian which still supports sysvinit as an option) so it won't be used anyway if there is a .service file.

Removing the old SysV init scripts is out of scope for this PR. We can clean up the codebase and remove them in subsequent PRs once we're sure they're not needed.

rpm/build/SPECS/jenkins.spec
basil: Can be done in a subsequent cleanup PR once we're sure they're not needed. You are welcome to submit that PR.

systemd/jenkins.service
basil: Fine, done in the latest commits

basil:
> where are the sources for […] https://pkg.jenkins.io/redhat-stable/jenkins.repo

I suggest running `git grep 'baseurl='` in a checkout of this repository.

basil:
Should fix most if not all of the following bugs:

- https://issues.jenkins.io/browse/JENKINS-31656
- https://issues.jenkins.io/browse/JENKINS-41218
- https://issues.jenkins.io/browse/JENKINS-46020
- https://issues.jenkins.io/browse/JENKINS-51888
- https://issues.jenkins.io/browse/JENKINS-54558
- https://issues.jenkins.io/browse/JENKINS-55053
- https://issues.jenkins.io/browse/JENKINS-55696
- https://issues.jenkins.io/browse/JENKINS-56219
- https://issues.jenkins.io/browse/JENKINS-57015
- https://issues.jenkins.io/browse/JENKINS-60763
- https://issues.jenkins.io/browse/JENKINS-63803
- https://issues.jenkins.io/browse/JENKINS-66277
- https://issues.jenkins.io/browse/JENKINS-67404
- https://issues.jenkins.io/browse/JENKINS-67487

basil:
Testers should take note that [PR build 33](https://ci.jenkins.io/job/Packaging/job/packaging/job/PR-266/33/) contains some quoting fixes in the migration script.

basil:
I've added yet another quoting fix to [PR build 35](https://ci.jenkins.io/job/Packaging/job/packaging/job/PR-266/35/). And let it be said: migrating the old Debian configuration to the new common shared format in pure `sh(1)` (as opposed to `bash(1)`) was absolutely brutal, especially because the old Debian configuration had fewer knobs and because pure `sh(1)` has such primitive string manipulation constructs. The end result is a 452-line monstrosity, but I have tested it with quite a few unquoted, single-quoted, and double-quoted strings for both Jenkins options and Java options, and all seems to be working fine now.

systemd/jenkins.service
```diff
@@ -0,0 +1,142 @@
...
+#     systemctl edit @@ARTIFACTNAME@@
+#
+# For more information about drop-in files, see:
+#
+#     https://www.freedesktop.org/software/systemd/man/systemd.unit.html
+#
+
+[Unit]
+Description=Jenkins Continuous Integration Server
+Requires=network.target
```
axmetishe, line 14: Dependency may be weaker - `Wants=`
https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Wants=

systemd/jenkins.service
```diff
@@ -0,0 +1,142 @@
...
+After=network.target
```
axmetishe, line 15: Better to use `After=network-online.target` to be sure that the network is ready

systemd/jenkins.service
```diff
@@ -0,0 +1,142 @@
...
+# Unix account that runs the Jenkins daemon
+# Be careful when you change this, as you need to update the permissions of
+# $JENKINS_HOME, $JENKINS_LOG, and (if you have already run Jenkins)
+# $JENKINS_WEBROOT.
+User=@@ARTIFACTNAME@@
+Group=@@ARTIFACTNAME@@
+
+# Directory where Jenkins stores its configuration and workspaces
+Environment="JENKINS_HOME=/var/lib/@@ARTIFACTNAME@@"
+WorkingDirectory=/var/lib/@@ARTIFACTNAME@@
```
axmetishe, line 39: %{workdir} macros can be used for definition or %h systemd specifier since user created have it at the same location according to pre-install scripts

systemd/jenkins.service
```diff
@@ -0,0 +1,142 @@
...
+
+# Unix account that runs the Jenkins daemon
+# Be careful when you change this, as you need to update the permissions of
+# $JENKINS_HOME, $JENKINS_LOG, and (if you have already run Jenkins)
+# $JENKINS_WEBROOT.
+User=@@ARTIFACTNAME@@
+Group=@@ARTIFACTNAME@@
+
+# Directory where Jenkins stores its configuration and workspaces
+Environment="JENKINS_HOME=/var/lib/@@ARTIFACTNAME@@"
```
axmetishe, line 38: `Environment="JENKINS_HOME=%h"`
since user created have it at the same location according to pre-install scripts

systemd/jenkins.service
```diff
@@ -0,0 +1,142 @@
...
+WorkingDirectory=/var/lib/@@ARTIFACTNAME@@
+
+# Location of the Jenkins WAR
+#Environment="JENKINS_WAR=/usr/share/java/@@ARTIFACTNAME@@.war"
```
axmetishe, line 42: `EnvironmentFile=-/etc/sysconfig/jenkins` can be used for optional parameters 
https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=

rpm - "%{_sysconfdir}/sysconfig/${SERVICE}"
deb - "%{_sysconfdir}/default/${SERVICE}"

https://docs.fedoraproject.org/en-US/packaging-guidelines/RPMMacros/
https://www.debian.org/doc/manuals/maint-guide/dother.en.html#initd

systemd/jenkins.service
```diff
@@ -0,0 +1,142 @@
...
+
+[Unit]
+Description=Jenkins Continuous Integration Server
+Requires=network.target
+After=network.target
+
+[Service]
+Type=notify
+NotifyAccess=main
+ExecStart=/usr/bin/@@ARTIFACTNAME@@
```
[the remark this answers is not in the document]
basil: > Wouldn't be a good idea

I do not see why it would be a good idea.

systemd/jenkins.service
basil: > Dependency may be weaker

It may indeed.

systemd/jenkins.service
basil: > Better to use `After=network-online.target` to be sure that the network is ready

Possibly.

systemd/jenkins.service
basil: > %{workdir} macros can be used

Can indeed!

systemd/jenkins.service
basil: > since user created have it at the same location

But this is not guaranteed.

systemd/jenkins.service
basil: > `EnvironmentFile=-/etc/sysconfig/jenkins` can be used for optional parameters

But this is not generally recommended.

systemd/jenkins.service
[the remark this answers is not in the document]
axmetishe: - Remove bash memory footprint
- Reduce attack vectors by jvm only isolation in service
- Get rid of `eval` instructions

systemd/jenkins.service
axmetishe: By reducing dependency requirements you will reduce service startup time

systemd/jenkins.service
axmetishe: Well, looks like that despite the fact you are requesting review and tedting you're not welcome to any reasonable suggestions, so why are you requesting them?:)

systemd/jenkins.service
axmetishe: Default package behavior is well known that means you may rely on it and in case someone wants to override it it will go with drop-ins or other customization

systemd/jenkins.service
axmetishe: Would you please give me the link on where I can find this recommendation on systemd man page?

systemd/jenkins.service
timja: > Well, looks like that despite the fact you are requesting review and tedting you're not welcome to any reasonable suggestions, so why are you requesting them?:)

Unless it's wrong at this stage testing is more important 😄

axmetishe: What I can see from the PR that I still have to fully override service configuration from the default package by configuration management software the same way that I did for the last 6 years starting from 2016, so what the point to add weak code?=)

axmetishe: I do agree that I'm a bad person because I used systemd units for Jenkins for 6 years at least but didn't provide a PR with that but at this point you got people attention and don't want to use their experience:)
Looks strange to me.

timja: Feel free to submit a PR afterwards, to make these minor changes. 

This is an initial v1 for systemd, not frozen in time.

serianox:
Didn't install the package, because we already have one production env on Debian 11 that was already on systemd. So instead I manually ran the migrate.sh script and installed the /usr/bin/jenkins script and jenkins.service file.

The systemd service and start script works, but somehow systemd doesn't detect that Jenkins finished starting, then kills it and restart it, indefinitely.

I'll need to investigate to see if this is not related to my setup...

serianox:
> It requires Jenkins 2.333, see https://issues.jenkins.io/browse/JENKINS-41218?focusedCommentId=421312&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-421312
> 
> https://www.jenkins.io/changelog/#v2.333
> 
> > Startup completion notification is available with systemd(1)

Awww thanks @timja! :)

I'm still effectively running LTS. Worked around using `TimeoutStartSec=infinity` for now, I will remove it during the upgrade to 2.333.

ekohl:
@serianox you need to change the type to `simple`. `notify` requires 2.333. That's a much better workaround than an infinite timeout. "Fun" fact, a colleague just submitted https://github.com/systemd/systemd/issues/22385 for a bug with `TimeoutStartSec=infinity`.

basil:
@axmetishe To channel some of [Kohsuke's energy](https://github.com/jenkinsci/jenkins/pull/2754#issuecomment-280964827):

> I appreciate your code review, but let's focus on important things here.

Or to channel the energy of the late [Roger Faulkner](https://www.usenix.org/memoriam-roger-faulkner):

> You think this is so easy? Why don't you try writing it? From scratch.

My main objective for this PR is not to cause a regression for the thousands of users with existing installations who will be running `sudo {apt-get,dnf,zypper} upgrade` and running these bits soon enough. Applying a configuration change of this magnitude to such a wide variety of legacy installations that are not even visible to me (let alone under my control) is a formidable challenge. The migration script alone is 452 lines long. This constraint, in my opinion, is the primary reason this issue has remained unresolved for the past five (5) years. What may work for one environment may not work in another. The Jenkins project enjoys a broad spectrum of adoption, but this comes with significant backward compatibility challenges. [Compatibility matters](https://www.jenkins.io/project/governance/#compatibility-matters), and this is a core value of the Jenkins project.

As @timja mentioned, this is a v1. There is room for improvement. People should feel free to propose new PRs with their ideas for improvement once this PR is merged. We now have CI build/test automation using Ansible, so this repository has become friendlier to first-time contributors. Yes, we can remove the old System V init scripts. Yes, we can optimize startup time with regard to networking. Yes, we can make the shell scripts squeaky clean or possibly eliminate them altogether. But we can't do any of this if we don't first ship this PR and migrate our existing install base to `systemd(1)`.

Let's cross that bridge first and polish everything up on the other side. Join me in doing so.

wwuck:
We are running multiple Jenkins LTS 2.319.2 instances on Devuan 4 and Debian 11 without SystemD.  Both systems are running on `sysvinit`.

I installed the .deb file from [build 36](https://ci.jenkins.io/job/Packaging/job/packaging/job/PR-266/36/artifact/target/debian/jenkins_2.333_all.deb)
During the install, the debconf dialog asked me if I would like to overwrite the configuration file at `/etc/default/jenkins` with the new copy from the .deb file.
I selected to keep the existing file with my customisations and this caused Jenkins to fail startup after the new package version was installed.
```console
Configuration file '/etc/default/jenkins'
 ==> Modified (by you or by a script) since installation.
 ==> Package distributor has shipped an updated version.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** jenkins (Y/I/N/O/D/Z) [default=N] ? n
Installing new version of config file /etc/init.d/jenkins ...
insserv: warning: script 'savecache' missing LSB tags
insserv: Default-Start undefined, assuming empty start runlevel(s) for script `savecache'
insserv: Default-Stop  undefined, assuming empty stop runlevel(s) for script `savecache'
Restarting Jenkins Automation Server: jenkinsCorrect java version found
Correct java version found
Setting up max open files limit to 8192
 failed!
invoke-rc.d: initscript jenkins, action "restart" failed.
dpkg: error processing package jenkins (--install):
 installed jenkins package post-installation script subprocess returned error exit status 1
Errors were encountered while processing:
 jenkins
```
I then did a quick edit of `/etc/default/jenkins` to fix the updated line for `JENKINS_WAR=`.
```diff
# location of the jenkins war file
-JENKINS_WAR=/usr/share/$NAME/$NAME.war
+JENKINS_WAR=/usr/share/java/$NAME.war
```
After this edit, Jenkins started successfully with `service jenkins start`.

> Yes, we can remove the old System V init scripts.

I would like to please request that this is not done in the future as we are surely not the only people running Jenkins on Debian without SystemD.

MarkEWaite:
> I installed the .deb file from [build 36](https://ci.jenkins.io/job/Packaging/job/packaging/job/PR-266/36/artifact/target/debian/jenkins_2.333_all.deb) During the install, the debconf dialog asked me if I would like to overwrite the configuration file at `/etc/default/jenkins` with the new copy from the .deb file. I selected to keep the existing file with my customisations and this caused Jenkins to fail startup after the new package version was installed.
> 
> ```
> Configuration file '/etc/default/jenkins'
>  ==> Modified (by you or by a script) since installation.
>  ==> Package distributor has shipped an updated version.
>    What would you like to do about it ?  Your options are:
>     Y or I  : install the package maintainer's version
>     N or O  : keep your currently-installed version
>       D     : show the differences between the versions
>       Z     : start a shell to examine the situation
>  The default action is to keep your current version.
> *** jenkins (Y/I/N/O/D/Z) [default=N] ? n
> Installing new version of config file /etc/init.d/jenkins ...
> insserv: warning: script 'savecache' missing LSB tags
> insserv: Default-Start undefined, assuming empty start runlevel(s) for script `savecache'
> insserv: Default-Stop  undefined, assuming empty stop runlevel(s) for script `savecache'
> Restarting Jenkins Automation Server: jenkinsCorrect java version found
> Correct java version found
> Setting up max open files limit to 8192
>  failed!
> invoke-rc.d: initscript jenkins, action "restart" failed.
> dpkg: error processing package jenkins (--install):
>  installed jenkins package post-installation script subprocess returned error exit status 1
> Errors were encountered while processing:
>  jenkins
> ```
> 
> I then did a quick edit of `/etc/default/jenkins` to fix the updated line for `JENKINS_WAR=`.
> 
> ```diff
> # location of the jenkins war file
> -JENKINS_WAR=/usr/share/$NAME/$NAME.war
> +JENKINS_WAR=/usr/share/java/$NAME.war
> ```
> 
> After this edit, Jenkins started successfully with `service jenkins start`.

Thanks for testing.  I performed a similar test on my Debian 11 machine that was configured behind a reverse proxy.  The Debian standard handling of the modified file `/etc/default/jenkins` worked the same for me as it did for you.  I applied my changes to the configuration file `/etc/default/jenkins` and Jenkins restarted.  In my case, there were more options that I needed to add (`--prefix`), but it was the same workflow.

I assume that modifying the `/etc/default/xxx` file is a common way to do things on Debian when using System V init.  I assume that the `override.conf` file from systemd configuration avoids that type of problem by not being modified by the installer.  Is that correct or is there a risk that future changes may require a similar conflict resolution process?

ekohl:
> > Yes, we can remove the old System V init scripts.
> 
> I would like to please request that this is not done in the future as we are surely not the only people running Jenkins on Debian without SystemD.

My suggestion was limited to RPM or even just Red Hat where there is no sysvinit anymore. You are correct that Debian does give the choice.

basil:
> I assume that modifying the /etc/default/xxx file is a common way to do things on Debian when using System V init. I assume that the override.conf file from systemd configuration avoids that type of problem by not being modified by the installer. Is that correct or is there a risk that future changes may require a similar conflict resolution process?

@MarkEWaite Correct, there are no hard conflicts requiring mechanical merge resolution in `systemd(1)` drop-ins because these are separate files from the main service unit configuration. Of course, there may be soft conflicts (e.g., a user may have customized an environment variable, the environment variable may get renamed, and the user's customization might remain under the old name and thus become ineffective).

MarkEWaite:
I've run a series of successful upgrade tests experimenting with:

* Jenkins 2.330 on Debian 11 with prefix, custom Java installation, and custom Java garbage collection arguments upgraded to Jenkins 2.233 packaged in build 36 of this pull request.  The `JAVA_HOME` environment variable was not visible in the Jenkins "Manage Jenkins" -> "System information" of Jenkins 2.330 when set with `--javaHome` in the `JENKINS_ARGS` of /etc/default/jenkins , but becomes visible after upgrade to Jenkins 2.333 packaged in build 36 when placed in the `JENKINS_OPTS` environment variable

I've also reviewed the documentation pull request and commented there.

MarkEWaite:
I ran an upgrade test on Debian 10 that started with Jenkins 2.331 customized to use /opt/jdk-11 as JAVA_HOME, port 9090, custom garbage collection arguments, and `/jenkins` as the prefix.

In order to use a custom Java, I set `JAVA_HOME` in `/etc/default/jenkins` and also had to set `PATH=$JAVA_HOME/bin:$PATH` in `/etc/default/jenkins`.  Those changes to JAVA_HOME and PATH were not preserved during the upgrade process, while all the other settings were preserved (including the JENKINS_OPTS setting of `--javaHome`).  I assume the `JAVA_HOME` technique that I used in the 2.331 configuration is non-standard and likely works more as a happy accident.  

When I followed the instructions in the systemd configuration file to set `JAVA_HOME`, it behaved exactly as I expected.  I'm running a Jenkins 2.333 on Debian 10 with custom Java installation, custom garbage collection settings, a non-standard port, and a URL prefix.

MarkEWaite:
I ran an upgrade test on my Red Hat Enterprise Linux 8.5 installation with a custom prefix, custom Java installation, and environment variable settings for configuration as code.  The `systemd` override file felt much easier to modify as needed than the `/etc/sysconfig/jenkins` file.

I'd like a little more time to use the list of fixed bugs as ideas for more exploration.  I think that my exploratory tests could be completed as early as end of day Monday.

It would be nice to have a blog post that describes the changes and the ways that systemd makes things easier and simpler than the System V init process.

MarkEWaite:
Tested and confirmed that the build from this pull request resolves

- [ ] https://issues.jenkins.io/browse/JENKINS-31656
- [ ] https://issues.jenkins.io/browse/JENKINS-41218
- [ ] https://issues.jenkins.io/browse/JENKINS-46020
- [ ] https://issues.jenkins.io/browse/JENKINS-51888
- [ ] https://issues.jenkins.io/browse/JENKINS-54558
- [ ] https://issues.jenkins.io/browse/JENKINS-55053
- [ ] https://issues.jenkins.io/browse/JENKINS-55696
- [ ] https://issues.jenkins.io/browse/JENKINS-56219
- [ ] https://issues.jenkins.io/browse/JENKINS-57015
- [ ] https://issues.jenkins.io/browse/JENKINS-60763
- [ ] https://issues.jenkins.io/browse/JENKINS-63803
- [ ] https://issues.jenkins.io/browse/JENKINS-66277
- [ ] https://issues.jenkins.io/browse/JENKINS-67404
- [x] https://issues.jenkins.io/browse/JENKINS-67487

The docs team will want to include those fixes in the weekly changelog that first delivers this pull request.

systemd/jenkins.sh
```diff
@@ -0,0 +1,191 @@
...
+	JENKINS_JAVA_CMD="$(command -v java)" || return "$?"
+}
+
+check_java_version() {
+	java_version=$("${JENKINS_JAVA_CMD}" -version 2>&1 |
+		sed -n ';s/.* version "\([0-9]\{2,\}\|[0-9]\.[0-9]\)\..*".*/\1/p;')
+
+	if [ -z "${java_version}" ]; then
+		return 1
+	elif [ "${java_version}" != "11" ] && [ "${java_version}" != "1.8" ]; then
```
MarkEWaite, line 42: This provides no way for the administrator to use the `--enable-future-java` option to allow them to experiment with Java 17.  Is that intentional?  I thought that the previous code allowed non-matching Java versions if enable-future-java was passed in

basil: > Is that intentional?

A rhetorical question if I ever saw one. Obviously you see a problem. Why don't you just be direct and say: "That seems wrong to me because … (insert reasoning here)". Note that _guessing_ about what the previous code supported (rather than _confirming_ through testing) is a claim, not an argument.

I haven't tested the old code with Java 17. This version check is lifted from the old Debian version. I doubt the old Debian version worked with Java 17 (see also #217).

The old RPM version had its own logic for Java version selection which IMHO tried to be way too clever about guessing the Java version. I have not retained the old RPM logic in the `systemd(1)` port. I don't think we should be in the business of trying to guess the best Java version. Rather I just use what's on the `PATH` and give the user the option to override it if necessary. Simple and straightforward.

Perhaps the version check from the old Debian version is overkill as well. I don't object to removing it. If it can be shown that the old RPM version worked with Java 17 and that it now doesn't work with the transition to `systemd(1)`, that is technically a regression and I will do a minimal fix in this PR. Otherwise I think Java 17 support for all platforms is best left to a future PR that covers the Java 17 issue comprehensively and systematically with testing on all platforms, `--add-opens`, etc.

MarkEWaite: No regression as far as I can tell.  You're correct that the Debian code absolutely does not work with Java 17.  I was wrong in my thinking that the previous code allowed non-matching Java versions.

I had to modify the `/etc/init.d/jenkins` file to allow Java 17 with Jenkins 2.331, then had to modify `/etc/default/jenkins` to place JDK 17 in the `PATH` and set `JAVA_HOME` for JDK-17 and add `--enable-future-java`  to the Jenkins arguments.  That was enough to allow the controller to start, but that's already modifying files that are not intended to be modified.

I don't think we need to change the version check in this pull request.  When further work is done on Java 17 support, it will need changes in several locations.

Sorry for the distraction of a comment without the investigation that could have proven the comment was not needed.

timja:
Marking as on hold till after Wednesday security release.

serianox:
> @serianox you need to change the type to `simple`. `notify` requires 2.333. That's a much better workaround than an infinite timeout. "Fun" fact, a colleague just submitted [systemd/systemd#22385](https://github.com/systemd/systemd/issues/22385) for a bug with `TimeoutStartSec=infinity`.

@ekohl Thanks. For the record, I think the correct type is `exec`; it seems to work, whereas `simple` indefinitely hang. (anwsering here in case someone is in the same predicament as I)

timja:
Last couple of runs failed CI FYI @basil

timja:
There's a CI fix in https://github.com/jenkinsci/packaging/pull/283 which is unrelated to this PR so will ship anyway

PR #284, opened 2022-02-16
Stop customizing `TimeoutStartSec` in tests

basil:
With the release of jenkinsci/jenkins#6237, it should no longer be necessary to customize `TimeoutStartSec` in common use cases, including this repository's tests.

### DIFF
--- a/molecule/default/verify.yml
+++ b/molecule/default/verify.yml
@@ -17,7 +17,7 @@
       dest: /etc/systemd/system/jenkins.service.d/override.conf
       content: |
         [Service]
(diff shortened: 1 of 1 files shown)


timja (approved):

basil:
> it should no longer be necessary to customize `TimeoutStartSec` in common use cases

It turns out that starting 16 instances of Jenkins in parallel on a 4-vCPU VM with 16 GiB of RAM is not a common use case. Let's see if we can get this test to work by tuning up `jenkins.model.Jenkins.extendTimeoutSeconds`.

PR #285, opened 2022-02-17
Support port numbers less than 1024 in systemd

basil:
## Background

See [JENKINS-41218 (comment)](https://issues.jenkins.io/browse/JENKINS-41218?focusedCommentId=421905&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-421905) (CC @zbynek).

## Problem

When the Jenkins port is set to a privileged port (port numbers less than 1024), Jenkins fails to start with a permissions error.

## Solution

Direct the user to enable the [`CAP_NET_BIND_SERVICE` capability](https://manpages.debian.org/bullseye/manpages/capabilities.7.en.html#CAP_NET_BIND_SERVICE) in such scenarios, and update the migration script to do so automatically when migrating such configurations to `systemd`.

## Testing Done

Set `HTTP_PORT=80` in the legacy `/etc/default/jenkins` file and deleted any `systemd` drop-ins, then installed the package from this PR. Verified that the drop-in created by the migration script contained the `CAP_NET_BIND_SERVICE` and that Jenkins started successfully on port 80 and could accept HTTP connections.

### DIFF
--- a/systemd/jenkins.service
+++ b/systemd/jenkins.service
@@ -61,12 +61,18 @@ Environment="JAVA_OPTS=-Djava.awt.headless=true"
 #Environment="JENKINS_LISTEN_ADDRESS="
 
 # Port to listen on for HTTP requests. Set to -1 to disable.
+# To be able to listen on privileged ports (port numbers less than 1024),
+# add the CAP_NET_BIND_SERVICE capability to the AmbientCapabilities
+# directive below.
 Environment="JENKINS_PORT=@@PORT@@"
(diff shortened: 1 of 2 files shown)


MarkEWaite (approved):
Very elegant solution!  Thanks so much!

timja (approved):

zbynek:
Thanks for the fast reply!

zbynek (approved):

PR #295, opened 2022-03-10
Prefer --httpPort from JENKINS_ARGS over HTTP_PORT

nomis:
If the Jenkins options file looks like this:
```
  HTTP_PORT=8080
  JENKINS_ARGS="--httpPort=8081"
```

Then it is 8081 from "--httpPort=8081" that should be used, not 8080 from $HTTP_PORT but the migration script prefers $HTTP_PORT.

It's not entirely clear why HTTP_PORT exists in the options file at all when it wasn't used for this purpose by any init script. The debian version would check if the port can be used but it won't pass it to Jenkins. Only when the example "--httpPort=$HTTP_PORT" is added (which is not part of the default JENKINS_ARGS) will it do anything.

Prefer the HTTP port from $JENKINS_ARGS over the one in $HTTP_PORT.

### DIFF
--- a/systemd/migrate.sh
+++ b/systemd/migrate.sh
@@ -219,10 +219,10 @@ read_old_options() {
 		NEW_JENKINS_LISTEN_ADDRESS="${JENKINS_LISTEN_ADDRESS}"
 	fi
 
-	if [ -n "${HTTP_PORT}" ] && [ "${HTTP_PORT}" -gt 0 ]; then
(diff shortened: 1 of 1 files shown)


basil:
@nomis Thank you very much for your analysis and contribution! I think the problem is legitimate, and your solution is also valid. But I think I have a simpler solution in #296 if you are interested in taking a look.

PR #320, opened 2022-06-22
Update description

daniel-beck:
Align this description with the blurb from https://www.jenkins.io/press/

If we want to change the "official" description of Jenkins, we should start there.


MarkEWaite (approved):

NotMyFault (approved):

branding/description-file
```diff
@@ -1,5 +1,3 @@
-Jenkins is an open source automation server which enables developers around the world to reliably automate their development lifecycle processes of all kinds, including build, document, test, package, stage, deployment, static analysis and many more.
+Jenkins is the leading open source automation server supported by a large and growing community of developers, testers, designers and other people interested in continuous integration, continuous delivery and modern software delivery practices. Built on the Java Virtual Machine (JVM), it provides more than 1,500 plugins that extend Jenkins to automate with practically any technology software delivery teams use. In 2019, Jenkins surpassed 200,000 known installations making it the most widely deployed automation server.
```
lemeurherve, line 1: Is there a more recent installations number than the one from 20219?

daniel-beck: We recently went over 300k.

> If we want to change the "official" description of Jenkins, we should start there.

still applies.

PR #332, opened 2022-09-06
Merge master branch into stable-2.361.1

NotMyFault:
Prep for 2.361.1


krisstern:
Thanks @NotMyFault

NotMyFault:
> Why was this PR needed?

It's part of the LTS release checklist. The LTS part expects that the stable branch is up-to-date with changes from the master branch.

basil:
Well sure but this kind of blind merging is precisely the type of thing that Wadeck was objecting to in another thread. There needs to be a justification for each backport.

timja:
> Well sure but this kind of blind merging is precisely the type of thing that Wadeck was objecting to in another thread. There needs to be a justification for each backport.

this is not a backport as such, we freeze the line at the start of each new LTS line (can be frozen earlier if say weekly has moved on with a breaking change we don't want to pull).
After the .1 release then yes changes should be backported on an as needed basis.

basil:
> we freeze the branch at the start of each new LTS line

Then why even create the branch earlier? Just create it at the time that you are freezing it. But this doesn't make sense to me, since the entire point of a stable branch is to freeze it _before_ the release to create a window of time during which stabilization can occur.

NotMyFault:
> Then why even create the branch earlier?

A legitimate question. The packaging branch isn't needed until you plan to trigger the LTS job, is it?
The branch is created if you initiate a new LTS line and use the [init-lts-line](https://github.com/jenkins-infra/backend-commit-history-parser/blob/master/bin/init-lts-line#L27) script.
If this happens 3 or 4 weeks before the LTS release takes place, master and stable branches easily diverge.

daniel-beck:
> Then why even create the branch earlier?

Might be a side effect from making the LTS decision earlier before the previous LTS line is even finished. Previously, the baseline decision (and start of backporting) was the same week as .3 release, which would be a natural time to start creating branches.

basil:
> Why do we need to pick up packaging changes in weeklies after the baseline? Selectively for infra reasons makes sense, as well as regular bugfix backports like we do in core, but wholesale everything?

Exactly, that is my point as well regarding stabilization. The primary purpose of branching _x_ number of days/weeks prior to a release is to avoid picking up potentially buggy commits on the stable branch, i.e. to create a stabilization period. Syncing indiscriminately during that time exposes the stable branch to new bugs/regressions, which defeats the point.

PR #328, opened 2022-09-02
[JENKINS-68564] adding sept. LTS to packaging

kmartens27:
Adding September LTS 2.361.1 (and java 11 requirement) to templates base & header.war

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in [GitHub](https://github.com/jenkinsci/packaging/pull/312) or [Jira](https://issues.jenkins.io/browse/JENKINS-68564)
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #319, opened 2022-06-14
Allow Windows users to use the `docker` image while having checked out the repo when using Windows.

gounthar:
I am working on Windows most of the time.
I do know this platform is not supported to develop with Jenkins.
The documentation of this repo recommends using  `docker-compose`, which I did, but I got some problems because `git` was making the text files compatible with Windows, but I was about to use them with a `docker` container running `Ubuntu`.
So... I got some problems with carriage returns and new lines (#316).
This `.gitattributes` file allowed me to checkout in another directory and fire the `docker-compose` command and then subsequent commands (like `./prep.sh`) with no more errors linked to line endings.
Thanks to @dduportal for suggesting it.

As for the tests, if you have access to a Windows machine, 
 * clone my `gitattributes` branch: `git clone --branch gitattributes https://github.com/gounthar/packaging.git`
 * go in the directory and launch `docker-compose run --rm packaging bash`
 * when in the container, issue a `./prep.sh`

If you have no error like this one:
```bash
jenkins@420db984a78b:/srv/releases/jenkins$ ./prep.sh
: invalid option
```

then you're good to go.

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #333, opened 2022-09-06
Revert "[JENKINS-68564] adding sept. LTS to packaging"

lemeurherve:
Partially reverts jenkinsci/packaging#328

This PR has removed nuget.exe and is breaking the current Windows weekly release, reverting partially this PR to retrieve it

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -176,8 +176,8 @@
             </p>
 
             <dl>
-              <dt>2.346.1 (June 2022) and newer</dt>
-              <dd>Java 8, Java 11, or Java 17</dd>
+              <dt>2.361.1 (September 2022) and newer</dt>
+              <dd>Java 11 or Java 17</dd>
 
               <dt>2.164.1 (March 2019) and newer</dt>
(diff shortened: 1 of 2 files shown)


NotMyFault:
Should be backported to 2.361.1 to not break tomorrow's LTS release, if this fixes windows packaging.

MarkEWaite (approved):
Sorry that we missed the removal of a binary in the prior pull request.  Makes sense that would cause problems.

timja (approved):
should probably be installed in the docker image?

NotMyFault (approved):

basil (approved):
Probably caused by #319 and fixed in #329. Sorry I missed that in code review. The backport to `stable-2.361` does not have this problem.

basil:
> The backport to stable-2.361 does not have this problem.

Well #330 did not have this problem. #332, which was proposed for reasons that remain unclear to me, did have this problem.

basil:
Thanks again @lemeurherve, this was much appreciated!

Issue #346, opened 2022-10-25
Build fails complaining about `x-jenkins-systemd-platform` 


NotMyFault:
```
13:17:34  + ANSIBLE_FORCE_COLOR=true molecule test
13:17:34  CRITICAL Failed to validate /home/jenkins/workspace/Packaging_packaging_master/molecule/default/molecule.yml
13:17:34  
13:17:34  ["Additional properties are not allowed ('x-jenkins-systemd-platform' was unexpected)"]
```

https://github.com/jenkinsci/packaging/blob/master/molecule/default/molecule.yml#L7

PR #347, opened 2022-11-12
Expand entity in test definition

MarkEWaite:
## Expand entity in test definition

A component of the test framework seems to no longer tolerate the entity definition that is used to prevent duplication.  Expand the entity definition in the file so that the tests run successfully again.

Fixes https://github.com/jenkinsci/packaging/issues/346

### DIFF
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -4,68 +4,129 @@ dependency:
 driver:
   name: docker
 
-x-jenkins-systemd-platform: &jenkins_systemd_platform_anchor
-  command: /sbin/init
-  volumes:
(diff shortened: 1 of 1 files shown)


NotMyFault (approved):
I was a bit hesitant to revert this, but if it's the only working way, we better to do.

MarkEWaite:
> I was a bit hesitant to revert this, but if it's the only working way, we better do so.

I agree that this technique is ugly and unhealthy.  If there is a better way to fix the CI builds on the master branch, I'd much prefer to use it.  Currently we've had a failing CI job on the master branch for almost a month.  I'd much rather have ugly code that passes the CI job than a failing CI job.

PR #350, opened 2022-11-29
Add servlet container tests

basil:
Verify that Jenkins launches to the login screen with a custom `JENKINS_HOME` on Tomcat 9 and WildFly 26, which are the latest versions that support the `javax.servlet` API that we use. This is not a statement of official support for alternative servlet containers, but this verifies that we do not accidentally regress de facto working functionality. To test this I ran `molecule test -s servlet` locally.


basil:
See also https://github.com/jenkins-infra/jenkins.io/pull/5747

NotMyFault (approved):
Thanks!

PR #351, opened 2022-11-29
Verify login page loads

basil:
While implementing #350 I noticed that the regular tests could be improved to assert that the login page actually loaded with the expected content.

### DIFF
--- a/molecule/default/verify.yml
+++ b/molecule/default/verify.yml
@@ -35,7 +35,6 @@
       state: started
   - uri:
       url: "http://127.0.0.1:8080/login"
-      status_code: [-1, 200]
       return_content: true
     register: result
     until: result.status == 200
--- a/molecule/default/verify.yml
+++ b/molecule/default/verify.yml
@@ -34,12 +34,17 @@
       name: jenkins
       state: started
   - uri:
-      url: "http://127.0.0.1:8080"
-      status_code: 403
+      url: "http://127.0.0.1:8080/login"
+      status_code: [-1, 200]
+      return_content: true
     register: result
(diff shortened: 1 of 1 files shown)


molecule/default/verify.yml
```diff
@@ -34,12 +34,17 @@
       name: jenkins
       state: started
   - uri:
-      url: "http://127.0.0.1:8080"
-      status_code: 403
+      url: "http://127.0.0.1:8080/login"
+      status_code: [-1, 200]
```
daniel-beck, line 38: Log spam prevention, or why consider `-1` to be "success"? Wasn't clear from reading Ansible docs.

daniel-beck (approved):

molecule/default/verify.yml
basil: https://ci.jenkins.io/job/Packaging/job/packaging/job/PR-350/1/consoleFull flaked with:

```
01:05:32  fatal: [wildfly-26]: FAILED! => {"attempts": 20, "changed": false, "content": "", "elapsed": 0, "msg": "Status code was -1 and not [200]: Request failed: <urlopen error [Errno 111] Connection refused>", "redirected": false, "status": -1, "url": "http://127.0.0.1:8080/jenkins/login"}
```

I thought that the `urlopen` error was preventing it from looping and trying again, which is why I added -1 to the list, but on reading the logs more closely I see that it did loop 20 times, so this isn't make it anymore or less flaky. I suppose the fact that the test passed the second time had nothing to do with my addition of `status_code` and that it's actually still flaky. I'll remove the `status_code` customization at the very least.

PR #353, opened 2022-11-30
Support newer versions of systemd

### LINKED ISSUE #352: Tests no longer run on systemd > 248
### Jenkins and plugins versions report

Jenkins 2.379

### What Operating System are you using (both controller, and any agents involved in the problem)?

Ubuntu 22.04.1 LTS x86_64
systemd 249 (249.11-0ubuntu3.6)

### Reproduction steps

```bash
$ docker-compose run --rm packaging bash
$ ./prep.sh
$ make package
$ exit
$ python3 -m venv venv
$ . venv/bin/activate
$ pip install -r requirements.txt
$ molecule test
```

### Expected Results

The tests pass. They actually do pass in the CI build (which must be using a VM with systemd 248 or earlier) and used to pass on my local machine a few months ago, but no longer pass on my local machine in recent weeks after I have upgraded the packages in my local Ubuntu installation to systemd 249.

### Actual Results

The tests fail to run `systemctl daemon-reload`. `dmesg(1)` reports:

```
[  977.256889] systemd[1]: Set hostname to <debian-11>.
[  977.258394] systemd[1]: Failed to create /init.scope control group: Read-only file system
[  977.258625] systemd[1]: Failed to allocate manager object: Read-only file system
[  977.258661] systemd[1]: Freezing execution.
```

### Anything else?

https://github.com/systemd/systemd/issues/19245 and https://github.com/moby/moby/issues/42275 go into the gory details. Apparently this was never "supported" to begin with, and the "official" way to do this is to use Podman instead as in gjhenrique/rpi_stuff@10cf31f013bba7e89a2422a62effbbb581818786, though this would likely require some additional work to get Podman set up in the VM image for the CI build. johanssone/ansible-unbound@ec3cbb8440109f13d33a057801c7b65bb1b58095 shows a way to do this with Docker and systemd > 248, but this might break the CI build unless the VM image for the CI build is also updated to systemd > 248 in lockstep. In other words I suspect that any solution to this problem involves working on the VM image used for the CI build.

basil:
See [Docker and systemd, getting rid of dreaded 'Failed to connect to bus' error](https://www.jeffgeerling.com/blog/2022/docker-and-systemd-getting-rid-dreaded-failed-connect-bus-error).

Note that this PR upgrades us to `molecule-docker` `HEAD`. Support for `cgroupns_mode` was only released in 2.1.0, but that release also contains a regression that prevents us from using it. The regression is fixed in `HEAD` but unreleased. So by consuming `HEAD` we get the regression fix as well as support for `cgroupns_mode`, which is needed for newer versions of `systemd`. When a new version of `molecule-docker` is released we can upgrade to it and stop consuming `HEAD`.

Testing done: PR CI build passes before and after this change. Local build on Ubuntu 22.04.1 LTS x86_64 and systemd 249 (249.11-0ubuntu3.6) fails before this PR and passes after this PR.

Closes #352

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 ansible
 docker
 jinja2
-molecule-docker==2.0.0
+git+https://github.com/ansible-community/molecule-docker.git@main#egg=molecule-docker
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -8,124 +8,141 @@ platforms:
   # deb
   - name: debian-10
     image: dokken/debian-10:latest
-    command: /sbin/init
+    override_command: false
     volumes:
       - ${MOLECULE_PROJECT_DIRECTORY}/target:/var/tmp/target
-      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+      - /sys/fs/cgroup:/sys/fs/cgroup:rw
+    cgroupns_mode: host
     privileged: true
   - name: debian-11
     image: dokken/debian-11:latest
-    command: /sbin/init
+    override_command: false
     volumes:
       - ${MOLECULE_PROJECT_DIRECTORY}/target:/var/tmp/target
-      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+      - /sys/fs/cgroup:/sys/fs/cgroup:rw
+    cgroupns_mode: host
     privileged: true
   - name: ubuntu-18-04
     image: dokken/ubuntu-18.04:latest
-    command: /sbin/init
+    override_command: false
     volumes:
       - ${MOLECULE_PROJECT_DIRECTORY}/target:/var/tmp/target
-      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+      - /sys/fs/cgroup:/sys/fs/cgroup:rw
+    cgroupns_mode: host
     privileged: true
   - name: ubuntu-20-04
     image: dokken/ubuntu-20.04:latest
-    command: /sbin/init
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
 ansible
 docker
 jinja2
-molecule-docker==2.0.0
+git+https://github.com/ansible-community/molecule-docker.git@main#egg=molecule-docker
```
timja, line 4: could we pin to a commit please? this is the latest commit on master currently:
```suggestion
git+https://github.com/ansible-community/molecule-docker.git@5b26cc6cedb59de813229fd8a5bed7eb194db230#egg=molecule-docker
```

basil: Pleas do not make requests of me without providing the reasoning behind the request. And no I am not going to track a specific commit when nothing is updating it automatically to a newer version. If someone set up Dependabot for this file then I would be fine with a deterministic build.

timja: The reason is for deterministic CI builds, molecule has had breakages in more than one version why track their main branch?

timja:
image

Could you not have waited for my response? Why not just merge immediately =/

basil:
> Could you not have waited for my response? Why not just merge immediately =/

I did not wait because the original request lacked reasoning and because I felt that pinning the dependencies in `requirements.txt` to deterministic versions was out of scope of this PR. I filed #354 to track pinning and automatically updating the dependencies in `requirements.txt`.

basil:
⇒ #355

PR #355, opened 2022-11-30
Deterministic dependencies

### LINKED ISSUE #354: Pin and automatically update Python dependencies

As was pointed out in #353 the dependencies in `requirements.txt` are not installed deterministically nor updated automatically. The principle of reproducible builds dictates that dependencies should be installed deterministically, and the principle of software maintenance dictates that these dependencies be updated periodically, ideally automatically.

### Upstream changes

_No response_

basil:
Fixes #354

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-ansible
-docker
-jinja2
-git+https://github.com/ansible-community/molecule-docker.git@main#egg=molecule-docker
+ansible==7.0.0
+jinja2==3.1.2
+# Pending the release of https://github.com/ansible-community/molecule-docker/pull/185
+git+https://github.com/ansible-community/molecule-docker.git@5b26cc6cedb59de813229fd8a5bed7eb194db230#egg=molecule-docker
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
-ansible
-docker
-jinja2
-git+https://github.com/ansible-community/molecule-docker.git@main#egg=molecule-docker
+ansible==6.6.0
+jinja2==3.1.2
+# Pending the release of https://github.com/ansible-community/molecule-docker/pull/185
+git+https://github.com/ansible-community/molecule-docker.git@5b26cc6cedb59de813229fd8a5bed7eb194db230#egg=molecule-docker
```
basil, line 4: Every dependency in this file is pinned, which is consistent. However, I doubt that Dependabot will be able to automatically update the last one automatically. But hopefully the situation of fetching from Git rather than PyPI is only temporary, pending the release of https://github.com/ansible-community/molecule-docker/pull/185.

jtnord (approved):

PR #361, opened 2023-02-16
Switch to jenkins-io components for navbar and footer

### LINKED ISSUE #360: Switch to jenkins-io-components for navbar and footer generation

Instead of defining the header and footer manually in https://github.com/jenkinsci/packaging/blob/master/templates/footer.html and https://github.com/jenkinsci/packaging/blob/master/templates/base.html, we should switch to [jenkins-io-components](https://github.com/jenkins-infra/jenkins-io-components) for generation, similar to jenkins.io, accounts.jenkins.io etc

### Upstream changes

_No response_

NotMyFault:
This PR switches to Gavin's jenkins-io components package, which we're already using on plugins.j, accounts.j, etc. to put a navbar and footer in place.

Fixes #360

Before:

Screenshot 2023-02-16 at 16 14 24
Screenshot 2023-02-16 at 16 20 54

After:

Screenshot 2023-02-16 at 16 19 18
Screenshot 2023-02-16 at 16 21 23


templates/footer.html
```diff
@@ -4,82 +4,7 @@
...
-                </ul>
-              </div>
-            </div>
-          </div>
-        </div>
-      </div>
-    </div>
-  </div>
-</footer>
+<jio-footer property="https://jenkins.io"/>
```
timja, line 7: should be www to prevent a redirect and to match www.jenkins.io already
```suggestion
<jio-footer property="https://www.jenkins.io"/>
```

templates/base.html
```diff
@@ -51,84 +51,12 @@
...
-                </div>
-            </li>
-            <li class='nav-item'>
-                <a class='nav-link btn btn-outline-secondary' href='https://www.jenkins.io/download'>Download</a>
-            </li>
-          </ul>
-        </div>
-      </div>
-    </nav>
+    <jio-navbar class="fixed-top" property="https://jenkins.io"/>
```
timja, line 59:
```suggestion
    <jio-navbar class="fixed-top" property="https://www.jenkins.io"/>
```

timja (approved):

PR #365, opened 2023-03-21
Fix styling of debian install instructions

NotMyFault:
Before:
Screenshot 2023-03-21 at 20 14 54

After:
Screenshot 2023-03-21 at 20 15 46

Always bothered me and kind of looks like that there's a line missing when it's not.


MarkEWaite (approved):
Thanks!

PR #367, opened 2023-03-21
Include jio components in 2.387 stable branch

NotMyFault:
I replaced the header and footer on get.jenkins.io with our jio components library, which we're using on all other *.jenkins.io pages, on February the 16th. Starting with the following weekly release, weekly pages use the new header and footer layout.

Does anyone have strong objections about including these changes in the current stable branch?
I would like to get it into `stable-2.387` to warrant style consistency across all pages.
Otherwise, the stable pages are stuck with the old style until May the 31st.


basil (approved):
Thanks!

MarkEWaite (approved):
+1 from me.  Thanks very much!

PR #379, opened 2023-03-27
Add a banner to RH/deb pages about GPG key changes

NotMyFault:
Ref https://github.com/jenkins-infra/helpdesk/issues/3457

Screenshot 2023-03-27 at 16 55 19

Thoughts?

### DIFF
--- a/templates/header.debian.html
+++ b/templates/header.debian.html
@@ -5,7 +5,7 @@
 <div class="alert alert-danger">
     <b>Important Notice:</b> Beginning with LTS 2.387.2 and weekly 2.397, releases will be signed with a new GPG key. <br>
     Administrators <b>must</b> install the new key on their servers <b>before</b> attempting to update Jenkins. <br>
-    <a target="_blank" href="https://jenkins.io/blog/2023/03/27/TODO/">Read more about the key rotation on the blog</a>.
+    <a target="_blank" href="https://jenkins.io/blog/2023/03/27/repository-signing-keys-changing/">Read more about the key rotation on the blog</a>.
 </div>
 
 <p>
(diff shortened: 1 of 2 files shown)


templates/header.debian.html
```diff
@@ -2,6 +2,12 @@
 
 {% block distribution_instruction %}
 
+<div class="alert alert-danger">
+    <b>Important Notice:</b> Beginning with LTS 2.387.2 and weekly 2.397, releases will be signed with an updated GPG key. <br>
```
dduportal, line 6:
```suggestion
    <b>Important Notice:</b> Beginning with LTS 2.387.2 and weekly 2.397, releases will be signed with a new GPG key. <br>
```

templates/header.redhat.html
```diff
@@ -2,6 +2,11 @@
 
 {% block distribution_instruction %}
 
+<div class="alert alert-danger">
+    <b>Important Notice:</b> Beginning with LTS 2.387.2 and weekly 2.397, releases will be signed with an updated GPG key. <br>
```
dduportal, line 6:
```suggestion
    <b>Important Notice:</b> Beginning with LTS 2.387.2 and weekly 2.397, releases will be signed with a new GPG key. <br>
```

dduportal (commented):
Non blocking proposal to materialize there is a new key version.

Also, we might want to add at least the new link if it is ok for you?

Something like this (as per the code below):

```
{{web_url}}/{{organization}}-2023.key
```

NotMyFault:
> Non blocking proposal to materialize there is a new key version.

Oic, I didn't check out the puppet PR yet. Nice idea with the version suffix.

MarkEWaite (approved):

MarkEWaite:
I assume that we can make this visible as soon as https://pkg.jenkins.io/debian/jenkins.io-2023.key and friends are visible to users.  Tuesday morning European time seems soon enough to me.

PR #380, opened 2023-03-27
Fix bootstrap js URL

NotMyFault:


NotMyFault:
When testing locally, I noticed we don't fetch any JS from bootstrap here. Hence, the broken link remained unnoticed. I'm gonna check, at a later point, whether we can drop it fully.

PR #384, opened 2023-04-03
Do not publish MSI

MarkEWaite:
## Do not publish MSI

MSI installer needs to be signed with a code signing certificate.  Our code signing certificate has expired.  Attorneys from the Linux Foundation and the Continuous Delivery Foundation are working to renew the code signing certificate. Once that is completed, these changes can be removed and we'll resume publishing the MSI installer for Windows.  

Exit early from make.ps1 in case it is used in the build process.

Matching pull request in the release repository:

* https://github.com/jenkins-infra/release/pull/356

### DIFF
--- a/make.ps1
+++ b/make.ps1
@@ -4,6 +4,11 @@ Param(
     [String] $target = "package"
 )
 
+# TODO: Remove when MSI code signing certificate is available
+# echo "Not packaging Windows components until code signing certificate is available"
+# echo "Exiting make.ps1 early"
+# exit 0
+
 # # refers to the definition of a release target
 # BRAND:=./branding/test.mk
 # include ${BRAND}
--- a/make.ps1
+++ b/make.ps1
@@ -5,9 +5,9 @@ Param(
 )
 
 # TODO: Remove when MSI code signing certificate is available
-# echo "Not packaging Windows components until code signing certificate is available"
-# echo "Exiting make.ps1 early"
-# exit 0
+echo "Not packaging Windows components until code signing certificate is available"
+echo "Exiting make.ps1 early"
+exit 0
 
 # # refers to the definition of a release target
 # BRAND:=./branding/test.mk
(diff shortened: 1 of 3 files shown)


make.ps1
```diff
@@ -4,6 +4,11 @@ Param(
     [String] $target = "package"
 )
 
+# TODO: Remove when MSI code signing certificate is available
+# echo "Not packaging Windows components until code signing certificate is available"
```
slide, line 8: The # is a comment character, so this wouldn't do anything.

make.ps1
```diff
@@ -4,6 +4,11 @@ Param(
     [String] $target = "package"
 )
 
+# TODO: Remove when MSI code signing certificate is available
+# echo "Not packaging Windows components until code signing certificate is available"
+# echo "Exiting make.ps1 early"
+# exit 0
```
slide, line 10:
```suggestion
echo "Not packaging Windows components until code signing certificate is available"
echo "Exiting make.ps1 early"
exit 0

```

make.ps1
MarkEWaite: Thanks!  Fixed in 984bdc5bba39cac9aae03cf2486914d32e6b1f03

basil (approved):
This PR seems unnecessary to me when you could simply set `WINDOWS_PACKAGING_ENABLED` to `false` in `jenkins-infra/release/Jenkinsfile.d/core/package` to prevent any of this code from being called. I will not block this PR though.

NotMyFault (approved):

timja:
Can be reverted I think the certificate has arrived

NotMyFault:
> Can be reverted I think the certificate has arrived

I can confirm, it has. Just needs to be updated on our end. Reverting means we need to put it in place by tomorrow noon, is that a realistic timeframe?

MarkEWaite:
> > Can be reverted I think the certificate has arrived
> 
> I can confirm, it has. Just needs to be updated on our end. Reverting means we need to put it in place by tomorrow noon (EU time), is that a realistic timeframe?

I don't think we can be done by tomorrow noon EU time.  I'd prefer to proceed with this in place for the weekly build so that we have the day tomorrow to implement.

PR #388, opened 2023-04-06
Upgrade Ansible from 6.7.0 to 7.4.0

basil:
Perhaps we can upgrade Ansible now that our agents are running on 22.04. `systemd` no longer seems to work in the Amazon Linux 2 Docker container when running on an Ubuntu 22.04 Docker host, so I have removed this test in favor of a test based on an Amazon Linux 2023 Docker container, which does work on an Ubuntu 22.04 Docker host. Without affecting our existing support for installing Jenkins on Amazon Linux 2, this allows us to continue to do automated testing on Amazon Linux and helps us keep our documentation up-to-date. Since Amazon Linux 2023 no longer ships a vanilla OpenJDK 11 package, I updted the test to use Corretto, updating the user-facing documentation in https://github.com/jenkins-infra/jenkins.io/pull/6245 as well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,8 +6,3 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-    ignore:
-      # Starting with 7.x, Ansible requires Python >=3.9, which our Ubuntu
-      # 20.04-based agent VMs do not provide.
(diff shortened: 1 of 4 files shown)


MarkEWaite (approved):
Heartily approved.  Nice to have Amazon Linux 2023 test.

PR #392, opened 2023-04-11
Update meta charset

NotMyFault:
Switch from the older syntax to the nowadays considered default.


MarkEWaite (approved):
https://stackoverflow.com/questions/4696499/meta-charset-utf-8-vs-meta-http-equiv-content-type says this is the preferred form.  Thanks

Issue #386, opened 2023-04-05
Generate pages independently of the release process


NotMyFault:
During the GPG update process, a few people have pointed out that the `-stable` pages have not been updated.

The current generation happens in [indexGenerator.py](https://github.com/jenkinsci/packaging/blob/master/bin/indexGenerator.py), if a weekly or LTS release takes place. If there is no release, the pages don't update.

I recommend separating the current progress to allow maintaining and updating the pages independent of core builds, to communicate future changes on the `-stable` pages too.

PR #202, opened 2020-12-02
Fast forward stable-2.263 to match master branch

MarkEWaite:
The stable-2.263 branch needs the changes that are on the master branch.

PR #402, opened 2023-05-31
Remove key rotation banner

NotMyFault:
The warning has been removed from the carousel at jenkins.io some time ago. I propose to remove it from this repository too.
The current LTS line, 2.401, will continue showing the warning, given it's the first new LTS line with the updated key.

### DIFF
--- a/templates/header.debian.html
+++ b/templates/header.debian.html
@@ -2,12 +2,6 @@
 
 {% block distribution_instruction %}
 
-<div class="alert alert-danger">
-    <b>Important Notice:</b> Beginning with LTS 2.387.2 and weekly 2.397, releases will be signed with a new GPG key. <br>
-    Administrators <b>must</b> install the new key on their servers <b>before</b> attempting to update Jenkins. <br>
-    <a target="_blank" href="https://jenkins.io/blog/2023/03/27/repository-signing-keys-changing/">Read more about the key rotation on the blog</a>.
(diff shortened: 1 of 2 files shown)


timja (commented):
won't these get updated on the next LTS point release?

basil:
Our current (informal) process followed in e.g. #202 has historically involved merging the main branch into stable branches. Introducing non-backportable PRs seems error-prone in that it would be all too easy for someone to follow past precedent without realizing that there is a non-backportable PR, especially if there is no way for a first-time release lead to identify such non-backportable PRs.

NotMyFault:
> Our current (informal) process followed in e.g. #202

This was almost 2 1/2 years ago. There's no item on the current LTS checklist that states that master should be merged into a stable-x branch. 
Everything goes through backporting from Jira (or GH issues, if applicable). Given I didn't file an issue to backport this PR, no action should be taken for 2.401.2/3(/4).

basil:
My point is that the current release scheme, in which stable branches are created from the main branch periodically, does not really support displaying different text for different branches (#386). If the text on `stable-2.387` intentionally diverged from the text on the main branch, that intentional divergence would be lost again when `stable-2.401` was created from the main branch just a few months later. The only functional purpose for stable branches in this repository at this time is to decrease the risk associated with last-minute code changes, not to display different text for difference branches.

PR #411, opened 2023-06-20
Prefer Java 17 in Windows installer

NotMyFault:
Does anyone feel strongly about not promoting Java 17?

This is just a tiny part, but Java 17 has been available in Jenkins for over a year by now. Therefore, this seems like a safe push.

### DIFF
--- a/msi/build/jenkins.wxs
+++ b/msi/build/jenkins.wxs
@@ -168,11 +168,11 @@
         Return="check"
         Impersonate="no" />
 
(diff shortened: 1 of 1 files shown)


timja (approved):

MarkEWaite:
Thanks very much.  The documentation will switch to recommend Java 17 very soon on all platforms.  This is a nice step to promote more Java 17 use.  

Debian 12 (bookworm) does not deliver a Java 11 package, so we need to either describe it as a documentation special case or we can switch the documentation to describe Java 17 instead of Java 11.  We're switching the documentation.

PR #414, opened 2023-06-26
Try Wildfly 27

NotMyFault:


basil (changes requested):
Based on the [Servlet Container Support Policy](https://www.jenkins.io/doc/administration/requirements/servlet-containers/), WildFly 27 or later is based on Servlet API 6 (Jakarta EE 10) or later with `jakarta.servlet` imports, which is unsupported by current weeklies. I do not see how this could work.

NotMyFault:
There are several posts on community.j, attempting to use Wildfly 27 and 28. I'm uncertain whether we have a servlet container policy somewhere, but this PR would be the proof that it can't work in the current state (obviously).

basil:
> I'm uncertain whether we have a servlet container policy somewhere

https://github.com/jenkinsci/packaging/pull/414#pullrequestreview-1499215867

PR #418, opened 2023-07-25
Fix RPM tests

basil:
Recent versions of AlmaLinux and Rocky Linux apparently allow you to install OpenJDK without installing the time zone data, without which Jenkins startup fails. Fix the test suite by ensuring the time zone data is installed for all RPM-based distributions before running the test.


MarkEWaite (approved):
Thanks!

PR #421, opened 2023-07-26
Use Java 17 for tests and documentation

basil:
Run tests on Java 17 (except on Debian 10, which does not have a Java 17 package) and refer to Java 17 in the documentation wherever possible.


MarkEWaite (approved):
Thanks!

PR #428, opened 2023-08-24
fix(systemd): Handles the JDK 21 beta versions

gounthar:
On certains platforms, Temurin's openJDK build outputs a version such as `openjdk version "21-beta" 2023-09-19`.
This is not recognized as a valid version in the `jenkins.sh` script, so Jenkins can't be properly installed/started.

I'm proposing an alternative version search, so the beta versions also work.

```bash
awk -F '"' '/version/ {print $2}' | awk -F '.' '{match($1, /^[0-9]+/); print substr($1, RSTART, RLENGTH)}'`
```

The `match($1, /^[0-9]+/)` part searches for the first sequence of digits at the beginning of the version string, and `substr($1, RSTART, RLENGTH)` extracts that matched portion, effectively giving us the major version number.
This should handle cases like "21-beta" and "17.0.7", and extract the correct major version number. 🤷 

### Testing done

I haven't been able to build on Debian WLS2 because of `/tmp/tmp.OFSjOSDlKO/debian/jenkinstest.dirs: 2: usr/share/jenkinstest: not found` but I tested directly various `java -version` outputs on several machines, and it worked each and every time.

### Submitter checklist
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- ~~Link to relevant issues in GitHub or Jira~~
- ~~Link to relevant pull requests, esp. upstream and downstream changes~~
- ~~Ensure you have provided tests - that demonstrates feature works or fixes the issue~~

### DIFF
--- a/systemd/jenkins.sh
+++ b/systemd/jenkins.sh
@@ -36,8 +36,8 @@ infer_java_cmd() {
 check_java_version() {
 	printf '%s' "${JENKINS_OPTS}" | grep -q '\--enable-future-java' && return 0
 
-	java_version=$("${JENKINS_JAVA_CMD}" -version 2>&1 |
-		sed -n ';s/.* version "\([0-9]\{2,\}\|[0-9]\.[0-9]\)\..*".*/\1/p;')
+  java_version=$("${JENKINS_JAVA_CMD}" -version 2>&1 |
+    sed -n ';s/.* version "\([0-9]*\|-beta\)\{1,\}".*/\1/p;')
 
 	if [ -z "${java_version}" ]; then
 		return 1
--- a/systemd/jenkins.sh
+++ b/systemd/jenkins.sh
@@ -36,8 +36,8 @@ infer_java_cmd() {
 check_java_version() {
 	printf '%s' "${JENKINS_OPTS}" | grep -q '\--enable-future-java' && return 0
 
-  java_version=$("${JENKINS_JAVA_CMD}" -version 2>&1 |
-    awk -F '"' '/version/ {print $2}' | awk -F '.' '{match($1, /^[0-9]+/); print substr($1, RSTART, RLENGTH)}'
+	java_version=$("${JENKINS_JAVA_CMD}" -version 2>&1 |
+    awk -F '"' '/version/ {print $2}' | awk -F '.' '{match($1, /^[0-9]+/); print substr($1, RSTART, RLENGTH)}')
 
 	if [ -z "${java_version}" ]; then
 		return 1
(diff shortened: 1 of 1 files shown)


timja (approved):
Works on Java 17 for me

basil:
This already works today by adding `--enable-future-java`, so this PR is effectively adding support for running without `--enable-future-java` on beta versions. To me, the current behavior of requiring a special argument for a beta version makes some degree of sense, because it is less supported than a final release. Do others think this behavior should be changed?

gounthar:
You're absolutely right, Basil. It functions well with the `--enable-future-java` flag, making it more a matter of choosing the desired behavior rather than a technical conundrum.

MarkEWaite:
I would rather not have Jenkins users running beta versions unless they provide the the `--enable-future-java` flag.  

However, I must acknowledge that Jenkins 2.420 no longer requires the `--enable-future-java` flag for JDK 21.  Jenkins 2.417 does require the flag for Java 21.

basil:
> I would rather not have Jenkins users running beta versions unless they provide the the `--enable-future-java` flag.

I concur.

> However, I must acknowledge that Jenkins 2.420 no longer requires the `--enable-future-java` flag for JDK 21 beta. Jenkins 2.417 does require the flag for Java 21 beta.

Yeah, the logic is inconsistent. We could:

- Make the logic in core require `--enable-future-java` for pre-release versions, making it consistent with the logic here (and also requiring us to make ATH pass in `--enable-future-java`, just as PCT already does)
- Make the logic here accept pre-release versions to be consistent with core
- Remove one copy of the logic since this is effectively duplicate code. My vote would be to remove the copy here, since the logic in core will run anyway, as long as you have at least Java 8 (because the logic in core, though written in Java, is compiled to Java 8 bytecode)

gounthar:
What looks strange to me is that Jenkins works well when started with a simple `java -war jenkins.war` with JDK21 Beta, and that it does not work with `systemd` on the very same machine.

gounthar:
> This already works today by adding `--enable-future-java`, so this PR is effectively adding support for running without `--enable-future-java` on beta versions.

It's only on the beta version for the JDK 21, not for all beta versions to come.

basil:
I don't have a strong preference, but my vote would be to remove any Java version checking logic from this repository and rely on the checks in core. The logic in this repository was useful in the past because core's version checking logic was compiled with too new bytecode, but now that core's version checking logic is compiled with Java 8 (i.e. old) bytecode the logic in this repository is redundant and less useful. Plus removing it would make it easier to maintain this repository.

timja:
I’d be in favour of that

basil:
Alternative proposed in #429. No strong feeling from me either way. Whatever we form consensus around, the other PR can be closed.

MarkEWaite:
I like the removal of the check more than the addition.  I think the addition has the added challenge that it creates a new dependency on `awk` that I'm not sure is guaranteed available in all our packaging.

MarkEWaite:
Closed this pull request after merge of

* #429 

Thanks for detecting the issue and proposing a solution!

PR #433, opened 2023-09-29
Remove unnecessary `tzdata-java` workaround

basil:
Part of the never-ending quest to remove temporary workarounds before they become permanent.


samrocketman:
Context for the curious https://github.com/jenkinsci/packaging/pull/418

PR #435, opened 2023-10-10
Download the Debian/Ubuntu repository signature with wget, not curl

SunBlack:
See https://github.com/jenkins-infra/jenkins.io/pull/6633. Seems there is currently no guide for the LTS version in this repo (maybe remove this install steps here and just add a link to [this](https://www.jenkins.io/doc/book/installing/linux/#debianubuntu))?

### DIFF
--- a/templates/header.debian.html
+++ b/templates/header.debian.html
@@ -5,12 +5,12 @@
 <p>
   This is the Debian package repository of {{product_name}} to automate installation and upgrade.
 
-  To use this repository, first add the key to your system:
+  To use this repository, first add the key to your system (for the Weekly Release Line):
 
   <pre class="text-white bg-dark">
     <code>
-  curl -fsSL <a href="{{web_url}}/{{organization}}-2023.key" style="color:white">{{web_url}}/{{organization}}-2023.key</a> | sudo tee \
-    /usr/share/keyrings/jenkins-keyring.asc > /dev/null</code>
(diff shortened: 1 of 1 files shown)


templates/header.debian.html
```diff
@@ -5,12 +5,12 @@
 <p>
   This is the Debian package repository of {{product_name}} to automate installation and upgrade.
 
-  To use this repository, first add the key to your system:
+  To use this repository, first add the key to your system (for the Weekly Release Line):
 
   <pre class="text-white bg-dark">
     <code>
-  curl -fsSL <a href="{{web_url}}/{{organization}}-2023.key" style="color:white">{{web_url}}/{{organization}}-2023.key</a> | sudo tee \
```
basil, line 12: Is the equivalent of `curl -sS` not `wget -nv`?

SunBlack: Indeed, just keep it in sync with the Fedora `wget` instruction [here](https://www.jenkins.io/doc/book/installing/linux/#fedora). I can change it, in case you want it, but in this case we should also adjust the Fedora part.

basil: +1 for being consistent with other pages.

NotMyFault (approved):

basil (approved):

MarkEWaite (approved):
Thanks!

Issue #458, opened 2024-04-01
Diagnostic message "no matches found: [signed-by=/usr/share/keyrings/jenkins-keyring.asc]" following installation instructions


hakre:
When following the (current) installation instructions on https://pkg.jenkins.io/debian/ a diagnostic message is given:

> no matches found: [signed-by=/usr/share/keyrings/jenkins-keyring.asc]

It is in the second installation step:

> Then add a Jenkins apt repository entry:
>
> ~~~ 
>   echo deb [signed-by=/usr/share/keyrings/jenkins-keyring.asc] \
>     https://pkg.jenkins.io/debian binary/ | sudo tee \
>     /etc/apt/sources.list.d/jenkins.list > /dev/null
> ~~~

The message stems from the shell expanding the glob expression (square brackets).

### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

Shell, Debian based system.

</details>

### What Operating System are you using (both controller, and any agents involved in the problem)?

Irrelevant, this is before operating Jenkins on the system.

### Reproduction steps

1. Follow the installation instructions on a Debian based system
2. Specifically the second one.

### Expected Results

Command should not expand glob expression in square brackets.

### Actual Results

Command expands glob expression in square brackets.

### Anything else?

Originally reported in https://community.jenkins.io/t/venue-for-issue-patch-on-installation-instructions/13851

The diagnostic message given irritates users that copy and paste the installation instructions unmodified into their shell and submit the command.

Best practices should be followed which includes quoting parameters to prevent unintended glob expansion. Many interactive shells have the option to expand globs enabled by default.

### Are you interested in contributing a fix?

Yes, already in preparation. [Mark Waite](https://community.jenkins.io/u/MarkEWaite) already referenced the two repositories to apply changes to.

hakre:
@MarkEWaite: Thanks for merging, while progressing I've now found out that https://pkg.jenkins.io/debian-stable/ is also affected. Will extend.

For

~~~
[DIR]	[debian-rc/](https://pkg.jenkins.io/debian-rc/)	2016-04-07 04:19	-	 
[DIR]	[debian-stable-rc/](https://pkg.jenkins.io/debian-stable-rc/)	2016-02-04 19:46	-
~~~

I see old dates, otherwise noticed instructions are with `sudo apt-key add -` and therefore do not have this same issue, however they may require streamlining as using `apt-key add` is not recommended any longer(?). No idea about that, so won't touch for an update until further notice.

[...]

PR #459, opened 2024-04-01
Quote to prevent glob expansion

hakre:
Issue: https://github.com/jenkinsci/packaging/issues/458

### Testing pending

Hopefully the CI pipeline generates the HTML and serves it so that it can be reviewed and tested.

```[tasklist]
### Submitter checklist
- [ ] Link to relevant pull requests, esp. upstream and downstream changes (unclear)
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue (limited ability)
```

### DIFF
--- a/templates/header.debian.html
+++ b/templates/header.debian.html
@@ -17,8 +17,8 @@
 
   <pre class="text-white bg-dark">
     <code>
-  echo deb [signed-by=/usr/share/keyrings/jenkins-keyring.asc] \
-    {{ web_url }} binary/ | sudo tee \
+  echo "deb [signed-by=/usr/share/keyrings/jenkins-keyring.asc]" \
+    "{{ web_url }}" binary/ | sudo tee \
     /etc/apt/sources.list.d/jenkins.list > /dev/null</code>
   </pre>
 </p>
--- a/templates/header.debian.html
+++ b/templates/header.debian.html
@@ -18,7 +18,7 @@
   <pre class="text-white bg-dark">
     <code>
   echo "deb [signed-by=/usr/share/keyrings/jenkins-keyring.asc]" \
-    "{{ web_url }}" binary/ | sudo tee \
+    {{ web_url }} binary/ | sudo tee \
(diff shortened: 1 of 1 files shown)


templates/header.debian.html
```diff
@@ -17,8 +17,8 @@
 
   <pre class="text-white bg-dark">
     <code>
-  echo deb [signed-by=/usr/share/keyrings/jenkins-keyring.asc] \
-    {{ web_url }} binary/ | sudo tee \
+  echo "deb [signed-by=/usr/share/keyrings/jenkins-keyring.asc]" \
+    "{{ web_url }}" binary/ | sudo tee \
```
MarkEWaite, line 21: Since web URLs are not expected to contain shell special characters, it seems safe enough to me to use the same format here as is used in the Jenkins install guide.

Optional, since both forms are valid.
```suggestion
    {{ web_url }} binary/ | sudo tee \
```

MarkEWaite (approved):

templates/header.debian.html
MarkEWaite: I've set myself a reminder to merge this in 24 hours.

MarkEWaite: Thanks.  I think that we should include the change so that it is consistent with the installation documentation.  If `web_url` ever includes a shell special character we'll deal with it at that time.  I believe that there are cases that even the surrounding double quotes won't solve.  Rather than be inconsistent, let's retain consistency between the two sites.

hakre: Late, I had this already on my box but didn't force push to update and then got distracted.

The only difference I had was in the message:

~~~
Quote to prevent glob expansion

NOTE: {{ web_url }} intentionally not quoted, a URL like
      https://pkg.jenkins.io/debian -or-
      https://pkg.jenkins.io/debian-stable is expected.

Issue: https://github.com/jenkinsci/packaging/issues/458
~~~

Just FYI @MarkEWaite

MarkEWaite: Thanks @hakre!

PR #465, opened 2024-05-27
chore(Jenkinsfile) inherit pod definition from existing pod template

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4106, this PR is a try to have an inherited pod template definition for the build agent on ci.jenkins.io.

The goal is to ensure that administrator-defined settings such as tolerations can be reused without having to maintain a full-fledged pod YAML.

I had difficulties to have the inheritance working with a YAML inline (of from file) so I chose t override the image in pure Groovy: the merge never happened which looks weird to me: either I got the inherited image or the local yaml but never the merge (tried implicit merge, or setting up all yaml merge strategies). I most probably missed something: another pair of eye would help.

Partial success as we have the expected image and environment variables, along with inherited pod. The agent starts but the test phase of the pipeline fails which might be related to the not-transplanted elements:

- SecurityContext (I can't find a way to specify it using Groovy function, only with rawYaml)
- Resources limits (it uses the 4vCPUs and 12 Gb from parent).


timja:
I've given you access, not sure if you'll have to create a new PR or not (or just replay)

dduportal:
OK, I'm definitievely missing something with the Kubernetes plugin:

- I can merge from groovy but not from yaml, even if I use the `merge()` method

=> Let's assume we want packaging to have it own pod template provided by jenkins-infra team

dduportal:
I got the following problem: if I use the `inheritFrom` directive for the `podTemplate()` step along with either `yaml` (inline YAML) or `read(.*)Yaml` (pod template read from a file, tried both trusted and untrusted), the merge does not seems to be performed deeply:
- Adding a new container template (e.g. NOT named `jnlp`) reuse the inherited pod definition and successfully adds the 2nd container
- But if the container template is named `jnlp`, either I got the child container only, losing parent Pod directives such as tolerations (with the `override()` default yaml merge method) or I only got the parent directive (if I switch to the `merge()` YAML merge method).

However, if I use Groovy form, then the merge is performed as expected, which makes me say it's not because of the parent pod template setup (default to `override()` for all and allows inheriting the yaml merge method).

cc @Vlatombe @jglick @timja if it rings a bell (I bet it's something I don't understand in the inheritance system).

=> As such, I'll stick to the groovy method for now, as a tradeoff between developer autonomy (this repository needs advanced tooling such as ansible which might require tuning the pod template) and "admin defined setup).

dduportal:
@timja @basil @MarkEWaite , this PR is ready to roll: it spins up a pod with succes from the `docker-packaging` image and the "Build" stage works as expected.

Please note that the tests errors are during the "Test" stage, and happens in a VM (Docker) agent.
The error is due to https://github.com/docker/docker-py/issues/3256 which is well described in https://github.com/ansible-collections/community.docker/issues/868. Which should be fixed with either Ansible 9.7.0 and/or Ansible 10 (and/or adding a custom collection for `docker-community`).

I leave you to it if you want to merge this PR despite the checks errors, or if you want to fix the Tests first?

timja (approved):
Thanks!

PR #486, opened 2024-07-19
Define `StartLimitBurst`

jglick:
Defines `StartLimitBurst` (and `StartLimitIntervalSec`) to prevent the Debian packaging from endlessly restarting a failing Jenkins service, as suggested by @basil in https://github.com/jenkinsci/jenkins/pull/9483#issuecomment-2237829184.

### Testing done

Built `make deb` using the Docker Compose environment based on a WAR built from https://github.com/jenkinsci/jenkins/pull/9483/commits/e92c1e9792a09939e052dcd8186efb240ff38242. Took me a while to figure out that you have to

```bash
gpg --import --batch credentials/sandbox.gpg
```

which was not really explained. On my host machine (unsure how to do this inside Docker, and not willing to start a VM) I installed the package, referring to https://www.jenkins.io/doc/book/installing/linux/#debianubuntu (I had not used a native package for Jenkins in many years), and set up Jenkins, installing `configuration-as-code`. Created `/var/lib/jenkins/jenkins.yaml`

```yaml
jenkins:
  systemMessage: 'hello from JCasC #1'
```

and used

```bash
sudo systemctl restart jenkins.service
```

to see the first message. Edited the config to

```yaml
jenkins:
  systemMessage: 'hello from JCasC #1'
  xxx: true
```

and restarted again; used

```bash
journalctl -u jenkins.service
```

to observe that the controller crashed, and the service keeps on restarting it endlessly. Referred to https://www.redhat.com/sysadmin/systemd-automate-recovery https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html https://www.freedesktop.org/software/systemd/man/latest/systemd.time.html to

```bash
sudo systemctl edit jenkins
```

resulting in `/etc/systemd/system/jenkins.service.d/override.conf`

```ini
[Unit]
StartLimitBurst=5
StartLimitIntervalSec=5m
```

and saw that now after five attempts it gives up. Fixed config

```yaml
jenkins:
  systemMessage: 'hello from JCasC #2'
```

and used

```bash
sudo systemctl reset-failed jenkins.service
sudo systemctl restart jenkins.service
```

and then it comes up with the second system message.

### DIFF
--- a/systemd/jenkins.service
+++ b/systemd/jenkins.service
@@ -13,6 +13,8 @@
 Description=Jenkins Continuous Integration Server
 Requires=network.target
 After=network.target
+StartLimitBurst=5
+StartLimitIntervalSec=5m
 
 [Service]
 Type=notify


basil (approved):
Thanks for the PR!

The `Limit*` options were moved from `Service` to `Unit` in `systemd` 229 (although both the old and new location are accepted by systemd 229 and up), and `StartLimitInterval` was renamed to `StartLimitIntervalSec` in `systemd` 230 (although both the old and new name are accepted by systemd 230 and up). We don't support Red Hat 7 anymore, Red Hat 8 has a new enough version of `systemd`, and automated tests pass on all our supported Linux distributions, so this PR is fine. I am just noting this for completeness.

> Took me a while to figure out that you have to `gpg --import --batch credentials/sandbox.gpg` which was not really explained.

I thought this was explained in this paragraph of the `README`:

> Run `./prep.sh` to perform the preparatory actions of downloading the WAR and importing the GPG key. Run `make package` to build all the native packages.

FTR it isn't necessary to build the package at all when testing changes to the `systemd` service unit, as you can use the prebuilt official package from https://www.jenkins.io/download/ and then just customize `JENKINS_WAR` to whatever WAR you want to test, along with any other `systemd` changes you want to test.

jglick:
> Run `./prep.sh` to perform the preparatory actions…

Ah, I must have missed this because I did not expect it to be needed when running in the Docker Compose environment.

PR #492, opened 2024-12-03
Add script to update Jenkins version info registry

slide:
Adds a script to the installer which will update the version information in the registry for Jenkins after an upgrade of the WAR file vie the Jenkins UI.

### Testing done

Tested by setting up a self-signed certificate to check the signing portion of build.ps1. Tested update script as part of of Jenkins community post: https://community.jenkins.io/t/windows-upgrading-in-place-doesnt-update-version-information-in-add-remove-programs/7069/7

### Submitter checklist

### DIFF
--- a/msi/build/Update-JenkinsVersion.ps1
+++ b/msi/build/Update-JenkinsVersion.ps1
@@ -1,0 +1,153 @@
+<#
+    .Synopsis
+    Upgrades the version information in the register from the current Jenkins war file.
+    .Description
+    The purpose of this script is to update the version of Jenkins in the registry
+    when the user may have upgraded the war file in place. The script probes the
+    registry for information about the Jenkins install (path to war, etc.) and 
+    then grabs the version information from the war to update the values in the
+    registry so they match the version of the war file. 
+
+    This will help with security scanners that look in the registry for versions
+    of software and flag things when they are too low. The information in the 
+    registry may be very old compared to what version of the war file is 
+    actually installed on the system.
+#>
+
+
+# Self-elevate the script if required
+if (-Not ([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole] 'Administrator')) {
+    # We may be running under powershell.exe or pwsh.exe, make sure we relaunch the same one.
+    $Executable = [System.Diagnostics.Process]::GetCurrentProcess().MainModule.FileName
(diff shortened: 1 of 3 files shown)


timja (approved):
I guess, people should really run the MSI though...

timja:
@kmartens27 FYI for changelog

timja:
> > I guess, people should really run the MSI though...
> 
> I am looking at disabling using the MSI for upgrades, it conflicts with the ability to upgrade the WAR through the Jenkins UI. We need to have one or the other, not both.

Its fairly common to be able to do both isn't it? An application can update itself, but if you run an installer it will also update it

Would be good if the upgrade process in Jenkins fully updated everything though.

slide:
> > > I guess, people should really run the MSI though...
> > 
> > 
> > I am looking at disabling using the MSI for upgrades, it conflicts with the ability to upgrade the WAR through the Jenkins UI. We need to have one or the other, not both.
> 
> Its fairly common to be able to do both isn't it? An application can update itself, but if you run an installer it will also update it
> 
> Would be good if the upgrade process in Jenkins fully updated everything though.

Usually in the cases where an application updates itself, it is not using an MSI, or downloads an updated MSI and runs that to upgrade. I agree that the Jenkins UI upgrade would be nice to include this update, but it requires modifying the registry, which requires admin access. If there is a way to run something as admin under Jenkins, we could do it.

slide:
No, there should not be. Is there any more details from the logs?

slide:
Feel free to revert to unblock the release

slide:
Ok, it looks like the `-Password` parameter for `Get-PfxCertificate` was added in Powershell 6.1, which it seems we don't have for this build. I will look at a different solution for signing (I think signtool can be used as well).

timja:
Or we use a newer powershell in the build?

dduportal:
> Or we use a newer powershell in the build?

Yeah, could be a good idea! I felt bad for having reverted your PR, and upgrading to a newer Powershell clearly is a good idea.

The packaging agent definition is a Windows pod template defined here: https://github.com/jenkins-infra/release/blob/master/PodTemplates.d/package-windows.yaml

Packages itself happens in https://github.com/jenkins-infra/release/blob/934635d1f0ab7c29be86f42f79461cdab70a6495/PodTemplates.d/package-windows.yaml#L31

timja:
I guess because its using dotnet framework 3.5 which is ancient?

https://github.com/jenkins-infra/release/blob/master/PodTemplates.d/package-windows.yaml#L31

PR #494, opened 2024-12-04
Add Powershell script to update registry information after war update

slide:
Adds a script to the Windows installer that can be used to update the registry information from the new war file after it is upgraded using the Jenkins UI. The script is signed using signtool which allows creating a function for the signing and calling it for the ps1 and the msi file(s) that are generated. Since the signing information is added to the ps1, it is first copied into tmp to avoid changing a version controlled file.

Supersedes #492 

### Testing done

I ran the signing enabled flow with a self signed certificate and was able to sign both the ps1 and msi.

### Submitter checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,5 @@ generated
 .idea/
 *.orig
 
-msi/build
 venv/
 
--- a/msi/build/Update-JenkinsVersion.ps1
+++ b/msi/build/Update-JenkinsVersion.ps1
@@ -1,0 +1,153 @@
+<#
+    .Synopsis
+    Upgrades the version information in the register from the current Jenkins war file.
+    .Description
+    The purpose of this script is to update the version of Jenkins in the registry
+    when the user may have upgraded the war file in place. The script probes the
+    registry for information about the Jenkins install (path to war, etc.) and 
+    then grabs the version information from the war to update the values in the
+    registry so they match the version of the war file. 
+
+    This will help with security scanners that look in the registry for versions
+    of software and flag things when they are too low. The information in the 
+    registry may be very old compared to what version of the war file is 
+    actually installed on the system.
+#>
+
+
+# Self-elevate the script if required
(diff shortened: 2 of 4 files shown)


slide:
I need to look at adding the msi build to the Jenkinsfile so we can catch things like the previous implementation before they get merged.

timja:
> I need to look at adding the msi build to the Jenkinsfile so we can catch things like the previous implementation before they get merged.

Did you want to do this before merging of this?

slide:
Let me scope it out at least

slide:
My scoping of the work to add the MSI build to the Jenkinsfile showed that it might be better to go to a newer version of Wix before trying that. Is there a way we can run a test release flow with this PR to make sure it won't break it?

timja:
> My scoping of the work to add the MSI build to the Jenkinsfile showed that it might be better to go to a newer version of Wix before trying that. Is there a way we can run a test release flow with this PR to make sure it won't break it?

I wonder if the easiest way would be to change it to the VM agent and then prepare it as part of adding CI to this repo?

ref https://github.com/jenkins-infra/helpdesk/issues/2745#issuecomment-1015167372

slide:
I think we can go ahead and merge this, unless anyone has any objections.

timja (approved):

timja:
cc @kmartens27 for changelog

kmartens27:
Thanks for the heads up @timja, should this be part of the weekly changelog for 2.512 or the next LTS (.3) changelog?

timja:
> Thanks for the heads up @timja, should this be part of the weekly changelog for 2.512 or the next LTS (.3) changelog?

Just weekly for now, it's not on the LTS branch: https://github.com/jenkins-infra/release/blob/stable-2.504/Jenkinsfile.d/core/package#L86

kmartens27:
Okay great, thanks very much for clearing that up! I'll be sure to include it in the weekly changelog this coming week.

PR #507, opened 2025-01-26
Remove some workarounds

basil:
Remove some workarounds that no longer seem necessary.


basil:
Seems to work fine except for Oracle Linux 8. @MarkEWaite How would you feel about dropping that from our test matrix?

MarkEWaite:
I think we can safely drop the testing of Oracle Linux 8.

MarkEWaite (approved):

Issue #503, opened 2025-01-26
Test on Fedora 41


basil:
Currently, we're testing Fedora 40 only when 41 is already available.

https://endoflife.date/fedora

SohamJuneja:
@basil I created a PR for this . However, the deployment tests on GitHub Jenkins failed. Could you confirm if my approach is correct? If needed, I can modify the PR or explore a different approach based on your feedback. Thanks!

PR #508, opened 2025-01-28
Test on Fedora 41

SohamJuneja:
This PR solves #503 by introducing support for Fedora 41 in the Molecule testing environment. The following changes have been made:

Updated molecule.yml: Added Fedora 41 as a platform in the test matrix for the Molecule configuration.
Verified Setup: Successfully pulled the fedora:41 Docker image and confirmed its functionality by running basic commands (uname -a, cat /etc/os-release) inside the container to ensure that Fedora 41 is set up correctly.
Test Validation: Ensured that Fedora 41 runs as expected in the Docker container, confirming that the setup is valid for testing purposes.
### Submitter checklist

### DIFF
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -79,7 +79,7 @@ platforms:
       - /sys/fs/cgroup:/sys/fs/cgroup:rw
     cgroupns_mode: host
     privileged: true
-  - name: fedora-41  # Update EOL when known
+  - name: fedora-41  # EOL 2025-11-19
     image: fedora:41
     override_command: false
     volumes:
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -79,6 +79,14 @@ platforms:
       - /sys/fs/cgroup:/sys/fs/cgroup:rw
     cgroupns_mode: host
     privileged: true
+  - name: fedora-41  # Update EOL when known
+    image: fedora:41
+    override_command: false
+    volumes:
+      - ${MOLECULE_PROJECT_DIRECTORY}/target:/var/tmp/target
+      - /sys/fs/cgroup:/sys/fs/cgroup:rw
+    cgroupns_mode: host
+    privileged: true
   - name: oraclelinux-9  # EOL 2032-06-30
     image: dokken/oraclelinux-9:latest
     override_command: false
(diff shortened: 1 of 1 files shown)


molecule/default/molecule.yml
```diff
@@ -79,6 +79,14 @@ platforms:
       - /sys/fs/cgroup:/sys/fs/cgroup:rw
     cgroupns_mode: host
     privileged: true
+  - name: fedora-41  # EOL 2025-11-19
+    image: fedora:41
```
MarkEWaite, line 83: I believe that the `dokken/fedora-41` container image is needed instead of the `fedora-41` container image.

The `fedora` container image is provided by the Fedora project and probably does not include all the things that are necessary for use with Molecule.
```suggestion
    image: dokken/fedora-41:latest
```

MarkEWaite (changes requested):

SohamJuneja:
@MarkEWaite Apologies for the oversight, and thank you for stepping in to make the necessary changes. Regarding the next steps, how would you suggest we proceed from here?

MarkEWaite:
> Regarding the next steps, how would you suggest we proceed from here?

The latest failures should be visible when you run the tests using the [instructions in the README](https://github.com/jenkinsci/packaging?tab=readme-ov-file#running-functional-tests).  If you can see those failures, then you can start diagnosing what might be causing those failures.

I'm out of office until next week, so won't be able to spend any time on it until next week at the earliest.

MarkEWaite:
> Regarding the next steps, how would you suggest we proceed from here?

The latest failures should be visible when you run the tests using the [instructions in the README](https://github.com/jenkinsci/packaging?tab=readme-ov-file#running-functional-tests).  If you can see those failures, then you can start diagnosing what might be causing those failures.

I'm out of office until next week, so won't be able to spend any time on it until next week at the earliest.

MarkEWaite:
> Regarding the next steps, how would you suggest we proceed from here?

The latest failures should be visible when you run the tests using the [instructions in the README](https://github.com/jenkinsci/packaging?tab=readme-ov-file#running-functional-tests).  If you can see those failures, then you can start diagnosing what might be causing those failures.

I'm out of office until next week, so won't be able to spend any time on it until next week at the earliest.

SohamJuneja:
@MarkEWaite Thanks for the update! I'll go through the test failures as per the README instructions and start diagnosing the issue

PR #530, opened 2025-08-20
Code cleanup of publish scripts

basil:
Fix quoting issues and use `[[` where possible (which is more idiomatic in bash scripts), as well as removing unnecessary calls to `echo`. These fix a number of `shellcheck` violations.

### Testing done

Verified that most `shellcheck` violations were gone after these changes. A few remain, but they are less easy to fix without thorough testing. On the other hand, the changes in this PR are low-hanging fruit and should be less risky.

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/deb/publish/publish.sh
+++ b/deb/publish/publish.sh
@@ -23,7 +23,7 @@ function generateSite() {
 	cp -R "$bin/contents/." "$D/contents"
 
 	gpg --export -a --output "$D/contents/${ORGANIZATION}.key" "${GPG_KEYNAME}"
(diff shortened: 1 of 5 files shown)


dduportal (approved):
Many thanks buddy! I restrained myself to fix these while working on it yesterday, this is a great addition!

PR #718, opened 2025-12-22
feat(deb,rpm) allow custom GPG public key file through a new environment variable 'GPG_PUBLIC_KEY_FILE' and fix missing Debian HTML

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4922

This PR ensures that the public GPG key used by the packaging process is:

- Published to the final webserver (`pkg.jenkins.io`) under the filename specified by the `GPG_PUBLIC_KEY_FILE` environment variable
  - It would have avoided the issue https://github.com/jenkins-infra/helpdesk/issues/3519 in 2023. And it will help us with the new 2026 key.
- Ensure the HTML pages for Debian and RPM are properly generated with the GPG key filename passed to the Python template engine
  - Includes a fix where the Debian HTML website was not uploaded to the correct location, so it was missing in our staging tests
  - Note: RPM HTML does not have a direct mention of the GPG key, but better to have it in case we change the documentation

Tested as part of https://github.com/jenkins-infra/release/pull/829 with success:

- The new key is published under the name `jenkins.io-2026.key`
- The HTML content for Debian is present and mentions the new key filename (and its new ID and expiration date)

### DIFF
--- a/bin/indexGenerator.py
+++ b/bin/indexGenerator.py
@@ -60,6 +60,7 @@ class IndexGenerator:
         self.product_name = os.getenv("PRODUCTNAME", "Jenkins")
         self.distribution = os.getenv("OS_FAMILY", "debian")
         self.gpg_pub_key_info_file = os.getenv("GPGPUBKEYINFO", ".")
+        self.gpg_public_key_filename = os.getenv("GPG_PUBLIC_KEY_FILENAME", "jenkins.io.key")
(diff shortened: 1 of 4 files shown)


lemeurherve (approved):
🚀

NotMyFault (approved):

timja (approved):
Not manually tested

PR #719, opened 2025-12-27
Update renovate config

NotMyFault:
As the title says.


MarkEWaite (approved):

MarkEWaite:
We are in process with several packaging changes related to the GPG signing key update and the RPM unification.  I'd prefer to not merge this for a week or two, while we focus on those higher priority items

PR #731, opened 2026-01-07
hotfix!: remove JDK17 from packaging (breaking)

lemeurherve:
This PR removes JDK 17 from packaging as 2.545 now requires JDK21: https://www.jenkins.io/changelog/2.545

Ref:
- https://github.com/jenkinsci/packaging/issues/729#issuecomment-3719079396

### Testing done

CI (molecule fails on Apple Silicon)

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

---

With this PR, we'll run a staging packaging to retrieve those new artifacts that we'll selectively redeploy to override the problematic 2.545 existing ones:

- MSI
- Debian: HTML only
- RPM all: HTML, RPM and index

See https://github.com/jenkinsci/packaging/issues/729#issuecomment-3719079396 for more details.

Issue #729, opened 2026-01-07
Jenkins 2.545 AlmaLinux package forces Java 17 dependency, breaks startup requiring Java 21


frankspiegel-coder:
**Describe the bug:**
After `dnf update` on AlmaLinux, Jenkins upgraded to 2.545-1.noarch but the package has a hard dependency pulling in `java-17-openjdk` (version 17.0.17.0.10-1.el9.alma.1). Jenkins now fails to start with "requires Java 21 or newer" due to PR jenkinsci/jenkins#26018 changes. Cannot remove Java 17 due to strong package dependency. Only solution is `dnf downgrade jenkins-2.544`.

**Expected behavior:**
Jenkins RPM should either:
1. Depend on `java-21-openjdk` instead of Java 17, or  
2. Have no Java dependency (user-provided as documented)

**System details:**
- AlmaLinux 9 (el9)
- Jenkins 2.545-1.noarch
- Triggered by PR jenkinsci/jenkins#26018 requiring Java 21+ [jenkinsci/jenkins#26018]
- Revert commit exists but package still broken [4c0689e]

**Steps to reproduce:**
1. `dnf install jenkins`
2. `dnf update` 
3. `systemctl start jenkins` → fails

**Workaround (unsustainable):**
`dnf downgrade jenkins-2.544`

**Packaging issue:** AlmaLinux Jenkins repo needs update to match Java 21 requirement from PR jenkinsci/jenkins#26018. Current package violates Jenkins support policy.

Related: jenkinsci/jenkins#26039 (revert issue)

ViliusS:
I have encountered similar RPM issue. Upgrading recent Jenkins pulls in 149 additional packages on Rocky Linux 10 which has no business being on the server:

<details>
```
Upgrading:
 docker-compose-plugin                                                        x86_64                                      5.0.1-1.el10                                                           docker-ce-stable                                      8.1 M
 jenkins                                                                      noarch                                      2.545-1                                                                jenkins                                                93 M
Installing dependencies:
 ModemManager-glib                                                            x86_64                                      1.22.0-7.el10                                                          baseos                                                324 k
 adwaita-cursor-theme                                                         noarch                                      46.0-3.el10                                                            appstream                                             521 k
 adwaita-icon-theme                                                           noarch                                      46.0-3.el10                                                            appstream                                             454 k
 at-spi2-atk                                                                  x86_64                                      2.56.1-1.el10                                                          appstream                                              87 k
[...]

PR #734, opened 2026-01-08
fix(rpm) define package release in a single location and set it back to `1`

dduportal:
This PR fixes our RPM tooling regarding the "build number" aka "Release" of the RPM packages.

- It defines a new environment variable `RPM_RELEASENUMBER` used to hold the RPP release. The goal is to avoid having the value set in 2 locations (which came as a surprise to us and required #733 after #731 when we wanted to increment this value to generate a new package).
  - The "Release" is used in both the RPM package filename (`jenkins-2.545-1.noarch.rpm`) and the RPM package metadatas (`rpm -qi jenkins-2.545-1.noarch.rpm | grep Release`).
  - Using an env. var allows changing the value only on the jenkins-infra/release when we need it, without having to perform commits in this repository
- The build number (default value) is set back to `1` instead of `2` (Partial rollback of #731 and #733)

It's part of the cleanup after https://github.com/jenkinsci/packaging/issues/729

----

Tests run:

- Locally with the Docker compose stack, I ensured we could
  - Build and publish the RPM with the default value `1` 
  - Build and publish the RPM with a custom value set to `3`
  - Build and publish the Debian package (no regression on the branding.py part):

```
$ docker compose up -d
# ... No errors ...
$ docker compose exec packaging bash
jenkins@c90b86daa9ed:/srv/releases/jenkins$ ./prep.sh
# ... No errors ...
jenkins@c90b86daa9ed:/srv/releases/jenkins$ make rpm.publish
# ... No errors ...
jenkins@c90b86daa9ed:/srv/releases/jenkins$ rpm -qi target/rpm/jenkins-2.545-1.noarch.rpm 2>/dev/null | grep Release
Release     : 1
jenkins@c90b86daa9ed:/srv/releases/jenkins$ RPM_RELEASENUMBER=3 make rpm.publish
# ... No errors ...
jenkins@c90b86daa9ed:/srv/releases/jenkins$ rpm -qi target/rpm/jenkins-2.545-3.noarch.rpm 2>/dev/null | grep Release
Release     : 3
jenkins@c90b86daa9ed:/srv/releases/jenkins$ make deb.publish
# ... No errors ...
```

- CI (in progress)


MarkEWaite (approved):

setup.mk
```diff
@@ -18,9 +18,10 @@ export MSI_SHASUM:=${MSI}.sha256
 # where to generate Debian/Ubuntu DEB file?
 export DEB=${TARGET}/debian/${ARTIFACTNAME}_${VERSION}_all.deb
 
-# TODO: restore package build number from `-2` to `-1` (hotfix of 2.545, see https://github.com/jenkinsci/packaging/issues/729)
+# What is the RPM Build Number ("Release" in the RPM spec, also known as "package suffix"). Can be overriden by setting the env. var 'RPM_RELEASENUMBER'.
```
lemeurherve:
```suggestion
# What is the RPM Build Number ("Release" in the RPM spec, also known as "package suffix"). Can be overriden by setting 'RPM_RELEASENUMBER' env. var .
```
(nit, not blocking)

lemeurherve (approved):
Tested locally with success, LGTM!

(Just had to specify a different release line, `2.528.1` in my case)

